PR #35, opened 2016-09-27
[Logging] Remove logging for typeahead

NiharikaRay:
Another one line fix - there's no reason to log an action that's not instantiated by the user, and that also clutters up the logs 


earthmancash:
Yup, definitely not a pageview.

PR #39, opened 2016-09-27
Bump version to 0.6.2.

matthewwardrop:


NiharikaRay:
One thing we may want to start doing is (in the PR description) put a note of the changes in the version the bump. This is a hotfix ({major}.{minor}.{hotfix}), so more necessary for when we do a major or minor bump, but let's do it for all.

NiharikaRay:
also the builds failed, I think because the branch was deleted before they could start the build.

PR #49, opened 2016-10-06
[Feature] Allow update endpoint

### LINKED ISSUE #48: knowledge repo content refresh
The only way I found at the moment to have a content up to date served by the webapp is to stop it and redeploy it. Is there a workaround to get a fresh content, like could we force the webapp to refresh the content every x hours?

NiharikaRay:
This is the original PR from @yolken to fix #48 

PR Additions: 
Creates an `/update` endpoint that users can hit to force an update of the index-data
Adds a setting in the server_config that sets how often the typeahead is refreshed

PTAL: @matthewwardrop @danfrankj @earthmancash

### DIFF
--- a/knowledge_repo/app/app.py
+++ b/knowledge_repo/app/app.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import
 import os
 import imp
 import logging
+import threading
 import traceback
 
 from flask import Flask, current_app, render_template, g, request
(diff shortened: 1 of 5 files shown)


earthmancash:
This LGTM, nice refactor with the added functionality. If we can add another endpoint to "pull latest git from all gitrepos", that'd make it so we wouldn't have to converge boxes at all, right?

matthewwardrop:
In case people come looking for this PR later, the patches in this PR are superseded by #105, which provides on-demand reindexing.

PR #50, opened 2016-10-06
[Feature] Make sure excluded tags are excluded everywhere

NiharikaRay:
While we have the concept of excluded tags, the tags and the posts weren't being excluded everywhere. This PR makes sure that all posts with excluded tags are hidden from everywhere. 

Test Plan: 
Add a "private" tag to the post and make sure it is not visible in any route

PTAL @matthewwardrop @danfrankj @earthmancash


earthmancash:
Nice, this LGTM. I think at some point we want a unit test for exclusion tags not showing up, but this is good.

Issue #72, opened 2016-10-12
running web app gets stuck in alembic migration


ametersky:
running `knowledge_repo --repo datasci_knowledge/ runserver` gives the following output.

I've been stuck on the last log for about ~10 mins.  I figure it shouldn't take this long because it's either an empty migration or it is initializing the db?

```
INFO:knowledge_repo.repositories.gitrepository:Fetching updates to the knowledge repository...
INFO:alembic.runtime.migration:Context impl SQLiteImpl.
INFO:alembic.runtime.migration:Will assume non-transactional DDL.
INFO  [alembic.runtime.migration] Context impl SQLiteImpl.
INFO  [alembic.runtime.migration] Will assume non-transactional DDL.
INFO  [alembic.runtime.migration] Running stamp_revision  -> 1b385158fd32
```

This is the page I am presented with at `localhost:7000`

![image](https://cloud.githubusercontent.com/assets/13753810/19322314/89c426a0-9086-11e6-8056-56b6e8ad39bf.png)

NiharikaRay:
Could you add the `--debug` flag to the command? That'll give some more information. 

`knowledge_repo --repo datasci_knowledge/ runserver --debug`

ametersky:
Adding the `--debug` flag doesn't give any extra output, unfortunately.

ametersky:
I just looked in `/tmp` to see if the database was ever created and I'm not seeing anything there.

ametersky:
Also, just realized this would show debug info in the web app, not the logs.  Here is the debug info from the web app :facepalm:

```
TypeError
TypeError: 'NoneType' object is not iterable

Traceback (most recent call last)
File "/Users/andrewmetersky/.virtualenvs/knowledge_repo/lib/python2.7/site-packages/flask/app.py", line 2000, in __call__
return self.wsgi_app(environ, start_response)
File "/Users/andrewmetersky/.virtualenvs/knowledge_repo/lib/python2.7/site-packages/flask/app.py", line 1991, in wsgi_app
response = self.make_response(self.handle_exception(e))
File "/Users/andrewmetersky/.virtualenvs/knowledge_repo/lib/python2.7/site-packages/flask/app.py", line 1567, in handle_exception
reraise(exc_type, exc_value, tb)
File "/Users/andrewmetersky/.virtualenvs/knowledge_repo/lib/python2.7/site-packages/flask/app.py", line 1988, in wsgi_app
response = self.full_dispatch_request()
[...]

PR #60, opened 2016-10-12
add Dockerfile

FvD:
To play around with knowledge-repo it would be helpful to have a Dockerfile. This one should work after #57 is solved. Right now knowledge-repo installs correctly, but I cannot `init` a local repo because a github ssh key is required.

PR #73, opened 2016-10-12
[Bugfix] Default excluded tags to empty

NiharikaRay:
This is the bug noted in https://github.com/airbnb/knowledge-repo/pull/60 and in https://github.com/airbnb/knowledge-repo/issues/72. 

If excluded tags was not in the server config, this would be None, and the for loop would fail. 

@matthewwardrop @earthmancash @danfrankj

### DIFF
--- a/knowledge_repo/app/app.py
+++ b/knowledge_repo/app/app.py
@@ -126,7 +126,7 @@ class KnowledgeFlask(Flask):
 
             # For every tag in the excluded tags, create the tag object if it doesn't exist
             # To ensure that posts with the excluded tags do not show up in the typeahead
-            excluded_tags = current_app.config.get('EXCLUDED_TAGS')
+            excluded_tags = current_app.config.get('EXCLUDED_TAGS', []) 
             for tag in excluded_tags:
                 tag_exists = (db_session.query(Tag)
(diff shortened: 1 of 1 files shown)


knowledge_repo/app/app.py
```diff
@@ -126,7 +126,7 @@ def update_typeahead_data(post):
 
             # For every tag in the excluded tags, create the tag object if it doesn't exist
             # To ensure that posts with the excluded tags do not show up in the typeahead
-            excluded_tags = current_app.config.get('EXCLUDED_TAGS')
+            excluded_tags = current_app.config.get('EXCLUDED_TAGS') or []
```
matthewwardrop: A safer way to do this is:
current_app.config.get('EXCLUDED_TAGS', [])
If the returned object is iterable or large then the way you've done it iterates through the entire list, and is a bit slower.

NiharikaRay: sounds good, I was initially going to do that but noticed elsewhere in the code you had defaulted to this paradigm 👍

matthewwardrop: My bad. Feel free to fix these up too :).

PR #79, opened 2016-10-15
Add Dockerfile

FvD:
Dockerfile to setup the knowledge-repo in a Docker container. To use this Dockerfile:

    $ docker build -t knowledgerepo .
    $ docker run --rm -p 7000:7000 --name airbnb -ti knowledgerepo 

This will run the container and land you on the command line, where you can set up a test repository and run the server as per the README file:

    $ knowledge_repo --repo ./example_repo init
    $ knowledge_repo --repo ./example_repo runserver

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ RUN apt-get -y install git
 RUN apt-get -y install ssh
 
 RUN pip install --upgrade pip
-RUN pip install knowledge-repo
+RUN pip install knowledge-repo[all]
 
 WORKDIR /knowledge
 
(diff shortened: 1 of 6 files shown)


FvD:
Maybe this should be closed without merging as well. I feel that again I created the pull request too early. This Dockerfile works to set things up, but only up to that initial point. There are still dependencies missing for things to work in R (for example `r-base` and common R packages like `knitr`, `rmarkdown` and `ggplot2`). 

As I work through the examples, and try to add my own Rmd and ipynb content, I realize this Dockerfile still needs work before it makes sense to submit it as a pull request. The idea would be to have one Dockerfile that allows for a generic deployment where at the very least all examples in the README work without issue.

PR #87, opened 2016-10-27
[Linguist] Fix language stats by excluding some js files

NiharikaRay:
@matthewwardrop @earthmancash @danfrankj 

2 minute fix that will hopefully stop the repo from being classified as a JS one based on the Linguist library that Github uses (https://github.com/github/linguist)


matthewwardrop:
Should we not exclude all of our external dependencies?

NiharikaRay:
Good catch - a better way of doing this was looking at `base.html` and figuring out what to exclude, which is what I did! PTAL

matthewwardrop:
Yep. Looks much better :).

PR #89, opened 2016-10-27
[Search] Add keyword search + ranking

NiharikaRay:
@matthewwardrop @danfrankj @earthmancash 

The much awaited search ranking PR. What this does: 
1. Actually fills in the keywords column, by combining the author, title, tldr, and tags of a post, and removing stopwords. 
2. Use the typeahead search to create a SQL statement where we rank the search results by the number of search terms that are present in the keywords of a post. Note: current limitation, right now, if a term like "nights" appears 4 times in one post, and 2 times in another post, those are considered equal. 
3. The typeahead front end is the same. 

Once this lands, I'll create a new tag and push out a release
@akuhn as an FYI

### DIFF
--- a/knowledge_repo/app/index.py
+++ b/knowledge_repo/app/index.py
@@ -3,6 +3,7 @@ import logging
 from .proxies import db_session, current_repo
 from .models import Post
 from .utils.emails import send_subscription_emails
+from .utils.search import get_keywords
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
@@ -28,6 +29,10 @@ def update_index():
             logger.info('Recording unpublished status for post at {}'.format(post.path))
             post.status = current_repo.PostStatus.UNPUBLISHED
             continue
+
+        # TODO(nikki_ray): This is to support the backfilling of this column. Remove this.
+        if not post.keywords:
+            post.keywords = get_keywords(post)
 
         # Update database according to current state of existing knowledge post and
         # remove from kp_dir. This means that when this loop finishes, kr_dir will
--- a/knowledge_repo/app/routes/render.py
+++ b/knowledge_repo/app/routes/render.py
@@ -2,6 +2,7 @@ import json
 import logging
 
 from flask import request, url_for, redirect, render_template, current_app, Blueprint, g
+from sqlalchemy import case, desc
 
 from ..proxies import db_session, current_repo
 from ..models import User, Post, PageView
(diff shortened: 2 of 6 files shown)


knowledge_repo/app/index.py
```diff
@@ -1,12 +1,17 @@
 from __future__ import absolute_import
 import logging
+
+from nltk.corpus import stopwords
```
matthewwardrop: This won't work if we have not run nltk.download(); which seems messy given our relatively mild dependence on NLTK. Can we just hard code the stopwords?

knowledge_repo/app/index.py
```diff
@@ -29,6 +34,8 @@ def update_index():
             post.status = current_repo.PostStatus.UNPUBLISHED
             continue
 
+        post.keywords = get_keywords(post)
```
earthmancash: @matthewwardrop one issue with running it only in the update_metadata_from_kp is that tag changing won't trigger a revision update. @NiharikaRay wonder if there's a place to call this for tag updating also?

knowledge_repo/app/routes/render.py
```diff
@@ -162,5 +163,28 @@ def about():
...
 @blueprint.route('/ajax_post_typeahead', methods=['GET', 'POST'])
-def ajax_post_typehead():
-    return json.dumps(current_app.config.get('typeahead_data', {}))
+def ajax_post_typeahead():
+    # this a string of the search term
+    search_terms = request.args.get('search', '')
+    search_terms = search_terms.split(" ")
+    case_statements = []
+    for term in search_terms:
+        case_stmt = case([(Post.keywords.like('%' + term.strip() + '%'), 1)], else_=0)
```
earthmancash: do we want to lowercase the term (and the keywords above)?

earthmancash:
Thanks for doing this Nikki!

knowledge_repo/app/routes/render.py
matthewwardrop: @earthmancash : Changing tags alone (without bumping post revision) already does not trigger an index update. Since tags are embedded in the document, checking every time is expensive. User contributed tags should be read separately, and applied after-the-fact, if we want to search by them at all.

NiharikaRay: I'll just change this to the `ilike` character, as opposed to the `like`.

knowledge_repo/app/index.py
NiharikaRay: Sure, I'll add this in a utils file so as to not clutter up the index.

knowledge_repo/app/index.py
earthmancash: That'd be great. I'd guess that others might prefer a workflow that tags posts after contribution as opposed to in the document. But that's still speculative, so can revisit that later if it's annoying to do now.

PR #93, opened 2016-10-28
[Search] Show title on mouse down + add background color

NiharikaRay:
The title is pretty self explanatory - on the new typeahead search, there was a regression where if you pressed the down arrow, you saw `[object object]` in the search box, as opposed to the title, so I fixed that + added a background color on the mouse down so you could see which suggestion you were currently at. 

Test Plan:
Write in the search box and make sure a title shows up + background color exists when you use the arrow keys to scroll

### DIFF
--- a/knowledge_repo/app/templates/base.html
+++ b/knowledge_repo/app/templates/base.html
@@ -114,7 +114,7 @@
         name: 'knowledge_posts',
         limit: 10,
         display: function (item) {
-          return item;
+          return item.title + " - " + item.author;
         },
         templates: {
           empty: Handlebars.compile(
--- a/knowledge_repo/app/static/css/custom.css
+++ b/knowledge_repo/app/static/css/custom.css
@@ -16,6 +16,10 @@
  cursor:pointer;
 }
 
+.tt-suggestion.tt-cursor {
+  background-color: #f5f5f5;
+}
+
 .tt-input, .tt-hint {
   width: 250px;
 }


knowledge_repo/app/templates/base.html
```diff
@@ -114,7 +114,7 @@
         name: 'knowledge_posts',
         limit: 10,
         display: function (item) {
-          return item;
+          return item.title;
```
earthmancash: Can we add author to this? I find it's usually the author/title combo that makes it clear to me what I'm looking at.

NiharikaRay: added - just as an aside though, most of the titles are longer than the width of the searchbox anyway, so the author won't show up.

knowledge_repo/app/templates/base.html
```diff
@@ -114,7 +114,7 @@
         name: 'knowledge_posts',
         limit: 10,
         display: function (item) {
-          return item;
+          return item.title + " - " + item.author;
```
earthmancash: can you have author before the title?

NiharikaRay: I think this is a more philosophical thing - I would assume seeing a title is more relevant than seeing a person's name, so I'd rather see as much of the title as possible than see a full name + parts of a title. 

2. A lot of posts have multiple authors, so then you'd only see the authors, which isn't super helpful. 

3. The suggestions are in the form `title - author`, so I think we should be consistent between the suggestion + what is displayed

earthmancash: Yeah I see now that you can still see the full highlight with title/author when toggling down on the options. This is fine.

PR #96, opened 2016-11-01
Improve logic around deployment of the knowledge repo.

matthewwardrop:
This patch set improves the deployment process, in part based on work by @yolken. Where possible, it does not write out temporary files but instead just launches gunicorn from within-process. This is useful when trying to compile knowledge_repo into a standalone executable format (such as pex).

This solves #97.

### DIFF
--- a/knowledge_repo/app/deploy/common.py
+++ b/knowledge_repo/app/deploy/common.py
@@ -1,0 +1,86 @@
+import os
+import types
+import inspect
+import textwrap
+from abc import abstractmethod
+from future.utils import with_metaclass
+
+from knowledge_repo import KnowledgeRepository
+from knowledge_repo.utils.registry import SubclassRegisteringABCMeta
+
+
+def get_app_builder(uris, debug, db_uri, config, **kwargs):
+    def get_app():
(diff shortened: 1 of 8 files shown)


knowledge_repo/app/deploy/common.py
```diff
@@ -0,0 +1,104 @@
...
+        # Repository uris will break as above (but less often since they are not often written too), but will also
+        # end up being a separate repository per thread; breaking consistency of presented content.
+
+        app = app or self.app
+
+        index_db = app.config['SQLALCHEMY_DATABASE_URI']
+        if index_db.startswith('sqlite://') and ':memory:' not in index_db:
+            return False
+
+        for uri in app.repository.uris.values():
```
danfrankj: if the app only has one repository, what is uris set to?

danfrankj:
one question about making sure we're supporting single repositories and then LGTM

knowledge_repo/app/deploy/common.py
matthewwardrop: The `uris` attribute is always a dictionary mapping "mountpoints" to urls. It was created with this kind of problem in mind :).

PR #100, opened 2016-11-01
added ability to have proxy posts

danfrankj:
one of my favorite one-line changes... 

Markdown files that have a proxy key are rendered as proxy posts

image

image


matthewwardrop:
@danfrankj This looks great. Since we don't have any security about user provided content yet, this looks fine. We might want to do something like [this](http://stackoverflow.com/questions/819416/adjust-width-height-of-iframe-to-fit-with-content-in-it) to automatically make the iframe match the height of embedded content. (Currently the iframe matches the height of the displayed area). Not sure if we want to do this in a separate PR?

Otherwise, LGTM.

danfrankj:
@matthewwardrop yeah let's leave that for another PR :)

PR #101, opened 2016-11-02
[Permissions] Add user permissions to YAML

NiharikaRay:
[Feature] This adds a field to the YAML header "users_allowed_to_view", which allows for special permissioning of posts. This is currently used to act as a "whitelist" for if a post has an excluded tag - ie, all users in this list of users can view a post with an excluded tag. 

In a future PR, I'll add a simple web view to add users to the permissions table, so that we don't need to clutter up git history.  I'll also add the concept of groups, so that posts can be shared with groups as opposed to just users. My plan was to get this PR (and release) out first so at least people could use the feature, and then add all the nice-to-haves.

Tested by editing a post, and then making sure I could see/not see the post depending on the user I was. 

@matthewwardrop @earthmancash @danfrankj PTAL


knowledge_repo/post.py
```diff
@@ -28,6 +28,7 @@
 - example
 created_at: 2016-06-29
 updated_at: 2016-06-30
+users_allowed_to_view: sally_smarts, wesly_wisdom
```
earthmancash: What do you think about leaving it out of the template for now since it's an edge case internally, and writing it up in the REAMDE tips & tricks? Thinking we might end up fielding a bunch of confused people on slack otherwise.

knowledge_repo/app/models.py
```diff
@@ -458,6 +460,19 @@ def update_metadata_from_kp(self, kp):
...
+            # comma separated list, add it into the Permission table
+            users_allowed_to_view = headers.get('users_allowed_to_view')
+            for user in users_allowed_to_view.split(","):
+                user = (db_session.query(User).filter(User.username == user.strip()).first())
+                if not user:
+                    user = User(username=user)
+                    db_session.add(user)
+                perm = Permission(user_id=user.id, post_uuid=self.uuid)
+                db_session.add(perm)
+                db_session.commit()
```
earthmancash: It looks like there's not a way to delete permissions. Can we make it so if the header isn't there, any permissions are deleted?

matthewwardrop:
Hi Nikki. Looking good.

Here's a few comments to do with as you will.

1) The yaml header name for permissions is a bit of a mouthful. I think having two headers "allowed_users" and "allowed_groups" is neater, almost as clear, and less of a typing marathon.
2) I think perhaps that the allowed_users and allowed_groups should be yaml lists, just like authors; and stored in an association table, again just like authors. The Permissions table you have is most of the way there. The list of permitted users and groups then becomes an attribute of the Post model, simplifying a lot of the rest of your code.
3) You shouldn't link to post uuid's as the id of a post in the index database. Refer instead to the post id, as everywhere else in the index. This maintains consistency of the index. Each post id will always point to the same UUID, unless a knowledge repository does not support UUIDs, in which case the post id will point to a null UUID.
4) Regarding future plans, I'm not convinced that effectively modifying post permissions server-side is a good idea; since if the post is ever updated in the future, it is unclear whether the server mandated or post mandated permissions should be used. It also gets the knowledge data repo out of sync. I think "groups" is the only authentication concept which should server mutable. If a post wants to mutate who can see the post post-publishing, it should subscribe to a group; not a static list of users. [The other place that this already happens is with post tags. User contributed tags should be similarly made distinct from author sanctioned tags; for much the same reasons].

I'm happy to help out here, so please don't feel like I'm dumping more work on you. Let me know your thoughts, and if there's any component of this you'd like to offload.

NiharikaRay:
Responding to your comments: 
1. Sure, that was what I first came up with. I can rephrase. 
2. I didn't build out the groups feature as a v0, since there's a user need for this right now. I can add that in a next PR. For the other stuff, sure, an association table is fine.
3. Sure, again, this seems like just semantics here. 
4. I vehemently disagree. This already happens with user tags, and the way we deal with this is that we always take the tags in the db, since when a post is first added, the tags get placed in the database. This is the same thing that should also happen with permissions. Furthermore, from a user experience side, as more people request access to a post, the author of a post shouldn't be forced to continually issue a github PR to update their posts, simply to add permissions.

knowledge_repo/post.py
NiharikaRay: sure! can do

earthmancash:
For 4.) I agree with Nikki that the tags concept for updating from post/server is a good analog for this. We should have the ability to change permissioning through the web app, perhaps only through the webapp if there's a good way to have it begin with permissions if needed.

matthewwardrop:
@earthmancash So you like that when a user updates their post, it resets all tags associated with the post, including those submitted by users? Seems a bit haphazard to me.

earthmancash:
@matthewwardrop I'd actually prefer that all tags are created through the webapp, and none are read through the post. It'd be great to do something similar with permissions, but I'm not sure there's a great way to make sure a sensitive post doesn't have the window between contribution and UI permissioning. I'm less concerned with post contribution resetting permissions since I'm guessing any permission changes on a post will go from more restrictive to less restrictive.

earthmancash:
@matthewwardrop as for now, I think our users would rather have a post resetting of metadata than to have to make PRs for every metadata switch.

matthewwardrop:
@earthmancash I am still not convinced :(. I think that being able to specify who can view it in the same flow as editing the post is a real boon, and that specifying allowed groups means we never need to arbitrarily change what is specified in the post itself. I'm not advocating that permissions should not be able to be changed on the server side, but rather, that changes on the server side should not *override* the permissions/tags set by the author (which should be in some sense privileged... since they were sanctioned by the author themselves).

It would be straightforward, for example, to store user contributed tags separately from author sanctioned ones, and to store permissions granted via the web interface separately as well. I think the key to permissions is groups. The author's original permissions should always been honoured, but augmented with the the set of groups managed by the knowledge repo app. These groups can include, for example, a 'post-level group', which can be modified by the app. i.e. 'projects/test.kp' is the name of a group that applies only to the projects/test.kp post. So any people listed in the original authors' post AND any speficied in the groups associated with the post (including the by-default post level group) can see the post.

NiharikaRay:
>I am still not convinced :(. I think that being able to specify who can view it in the same flow as editing the post is a real boon
This is correct, I don't think anyone is disagreeing with this.
> and that specifying allowed groups means we never need to arbitrarily change what is specified in the post itself. 
This is a false assumption
>The author's original permissions should always been honoured, but augmented with the the set of groups managed by the knowledge repo app. These groups can include, for example, a 'post-level group', which can be modified by the app. i.e. 'projects/test.kp' is the name of a group that applies only to the projects/test.kp post. So any people listed in the original authors' post AND any speficied in the groups associated with the post (including the by-default post level group) can see the post.

My question is - what if the author's original permission group changes? If, for example, author a decides to put person b, c, and d as viewers of the post. Then, server side, changes this to remove person c as a viewer. This would cause a contradiction, and in your case, then mean that person c is still able to view the post. Your philosophy works well when you need to augment users, but not when you need to take away permissions, which is an equally valid use case. 

My philosophy on this is to simplify. There's no reason to have two different "classifications" of permissions, one of which is defined by the author and one that is defined server-side (mind you, we can also make it so that server-side permissions can only be changed by admins + author), and then have the author's be ranked higher than the server's. Instead, we combine the two. This is how the flow works: 

1. A user only adds permissions to the YAML file -> great, this goes into the permissions table 
2. A user adds/delete permissions server-side after already having permissions in the YAML file -> we do the appropriate addition/deletion
3. A user adds/deletes in the YAML file, after already making server side changes -> if it's all stored in one place, we can easily do this operation as well. 

Then, when actually rendering the post, we only need to pull the permissions from one place.

matthewwardrop:
@NiharikaRay 

Thanks for the detailed rundown. :). With permissions at least, I can see why this makes sense. Permissions are only ever enforced through the app, and so it makes sense that the permissions are easily configurable on the server side  (and less so through the repository).

There is still some potential security issues; such as if a person is removed from the yaml headers, because as soon as that happens, it is not clear whether that person was originally added through the web ui... but I guess you could manage that during the actual index update.

Okay... I'm leaning more toward what @earthmancash suggested long term. Maybe we keep the 'allowed_groups' idea, and set a yaml flag of 'start_published' or some such; which sets the initial condition of whether the post should be revealed immediately, and then allows the author to go to the web app and manually specify particular people to be added to the post.

knowledge_repo/postprocessors/format_checks.py
```diff
@@ -13,7 +13,8 @@
 OPTIONAL_FIELD_TYPES = {
     'path': str,
     'updated_at': datetime.datetime,
-    'users_allowed_to_view': str,
+    'start_publishing': bool,   # If false, this post starts out private
```
matthewwardrop: I thought we were going for `private` rather than `start_publishing`

NiharikaRay: 👍

knowledge_repo/postprocessors/format_checks.py
```diff
@@ -12,7 +12,9 @@
 
 OPTIONAL_FIELD_TYPES = {
     'path': str,
-    'updated_at': datetime.datetime
+    'updated_at': datetime.datetime,
+    'start_publishing': bool,   # If false, this post starts out private
```
matthewwardrop: I thought we were going for `private` rather than `start_publishing`. Did something change?

knowledge_repo/app/models.py
```diff
@@ -247,6 +261,8 @@ class Post(db.Model):
     keywords = db.Column(db.Text)
     thumbnail = db.Column(db.Text())
 
+    start_publishing = db.Column(db.Integer())
```
matthewwardrop: Are we going for `private` instead of `start_publishing` (or `start_published`)? I think its cleaner.

PR #104, opened 2016-11-02
refactor of webeditor code

danfrankj:

Refactor of webeditor flow including

 * use current_repo rather than Post() models to retrieve and manipulate posts 
 * allow users to specify which repo to submit a post to 
 * various cleanup 
 * security updates so you can't delete others posts / comments 

image

TODOs 
* a few cases where we do need the post id to retrieve comments, etc. 
* testing

### DIFF
--- a/knowledge_repo/app/index.py
+++ b/knowledge_repo/app/index.py
@@ -40,7 +40,7 @@ def update_index():
         kp = kr_dir.pop(post.path)
 
         # Update metadata of post if required
-        if (kp.revision > post.revision or not post.is_published or kp.uuid != post.uuid):
+        if (post.revision is None or kp.revision > post.revision or not post.is_published or kp.uuid != post.uuid):
             if kp.is_valid():
                 logger.info('Recording update to post at: {}'.format(kp.path))
                 post.update_metadata_from_kp(kp)
--- a/knowledge_repo/app/routes/web_editor.py
+++ b/knowledge_repo/app/routes/web_editor.py
@@ -7,7 +7,7 @@ import os
 from collections import defaultdict
 from datetime import datetime
 from flask import request, render_template, Blueprint, current_app, url_for, send_from_directory, g
-from sqlalchemy import or_
+from sqlalchemy import or_, distinct
 from werkzeug import secure_filename
 
 from knowledge_repo.post import KnowledgePost
@@ -16,6 +16,7 @@ from ..models import Post, PostAuthorAssoc, Tag, Comment, User, PageView
 from ..utils.emails import send_review_email, send_reviewer_request_email
 from ..utils.image import pdf_page_to_png, is_pdf, is_allowed_image_format
 
+from ..index import update_index
 
 if sys.version_info.major > 2:
     from urllib.parse import unquote as urlunquote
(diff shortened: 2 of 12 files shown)


NiharikaRay:
> security updates so you can't delete others posts / comments
 LOVE IT THANK YOU.

knowledge_repo/app/routes/web_editor.py
```diff
@@ -88,228 +66,176 @@ def gitless_drafts():
...
-        else:
-            post = Post()
-            post.path = path
-    else:
-        path = post.path
-
-    # create the knowledge post
-    kp = KnowledgePost(path=path)
+def delete_post():
+    """ Delete a post """
```
NiharikaRay: it looks like a lot of these routes are exactly the same, can we consolidate it all into one function?

knowledge_repo/app/routes/web_editor.py
```diff
@@ -88,228 +66,176 @@ def gitless_drafts():
...
+    path = request.args.get('path', None)
+    if path not in current_repo:
+        return json.dumps({'msg': "Unable to retrieve post with path = {}!".format(path), 'success': False})
+    if g.user.username in current_repo.post(path).headers['authors']:
+        current_repo.remove(path)
```
NiharikaRay: If a person tries to remove a post that is not theirs, should we send back a  warning that tells them why the action failed?

knowledge_repo/app/routes/web_editor.py
```diff
@@ -35,43 +35,21 @@
 
 @blueprint.route('/ajax_tags_typeahead', methods=['GET'])
 def generate_tags_typeahead():
-    tag_objs = (db_session.query(Tag).all())
-    tags_typeahead = [str(tag_obj.name) for tag_obj in tag_objs]
-    return json.dumps(tags_typeahead)
+    return json.dumps([t[0] for t in db_session.query(Tag.name).all()])
```
danfrankj: Yeah I don't like the t[0] either, but this is in fact more efficient, given that the query only pulls the name, so I think we should keep it as is?

knowledge_repo/app/routes/web_editor.py
```diff
@@ -35,43 +35,21 @@
...
 
 
 @blueprint.route('/ajax_users_typeahead', methods=['GET'])
 def generate_users_typeahead():
-    user_objs = (db_session.query(User).all())
-    users_typeahead = [str(user_obj.username) for user_obj in user_objs]
-    return json.dumps(users_typeahead)
+    return json.dumps([u[0] for u in db_session.query(User.username).all()])
```
danfrankj: See above

knowledge_repo/app/routes/web_editor.py
```diff
@@ -88,228 +66,176 @@ def gitless_drafts():
...
-            can_approve = 1
-
-    tags_str = ''
-    if 'tags' in headers:
-        tags_str = ', '.join(headers['tags'])
-
-    # get created at and updated at, or default to now
-    created_at = (headers['created_at'] or datetime.now()).strftime('%Y-%m-%d')
-    updated_at = (headers['updated_at'] or datetime.now()).strftime('%Y-%m-%d')
+        data.update(headers)
```
danfrankj: Should be fairly easy :) will address in another PR

knowledge_repo/app/routes/web_editor.py
```diff
@@ -88,228 +66,176 @@ def gitless_drafts():
...
+            'repositories': current_repo.uri.keys(),
+            'markdown': None,
+            'comments': None,
+            'thumbnail': '',
+            'can_approve': 0,
+            'username': g.user.username,
+            'created_at': datetime.now(),
+            'updated_at': datetime.now()}
+
+    if path is not None and path in current_repo:
```
danfrankj: I'd like to be more explicit in this case unless you have strong reason not to be?

knowledge_repo/app/routes/web_editor.py
```diff
@@ -88,228 +66,176 @@ def gitless_drafts():
...
-    post = (db_session.query(Post)
-                      .filter(Post.id == post_id)
-                      .first())
-    post.status = current_repo.PostStatus.UNPUBLISHED
-    db_session.commit()
-
-    path = post.path
+    """ Unpublish the post """
+    path = request.args.get('path', None)
+    if not path not in current_repo:
```
danfrankj: yup fixed!

knowledge_repo/app/routes/web_editor.py
danfrankj: not sure exactly what you mean here?

knowledge_repo/app/routes/web_editor.py
danfrankj: done!

danfrankj:
@NiharikaRay @matthewwardrop @earthmancash this is no longer a WIP - ready for your full review :)

knowledge_repo/app/index.py
```diff
@@ -40,7 +40,7 @@ def update_index():
         kp = kr_dir.pop(post.path)
 
         # Update metadata of post if required
-        if (kp.revision > post.revision or not post.is_published or kp.uuid != post.uuid):
+        if (post.revision is None or kp.revision > post.revision or not post.is_published or kp.uuid != post.uuid):
```
matthewwardrop: @danfrankj Why is this necessary?

matthewwardrop:
Hey @danfrankj. This looks like a big improvement. Great work.

knowledge_repo/app/routes/web_editor.py
NiharikaRay: so you could have multiple routes point to the same function - something like 
```
@blueprint.route('/delete_post')
@blueprint.route('/accept_post')
def do_the_function()
    return "ok"
```
The reason I suggested this was because for most of the functions you seem to be getting the path, checking to see if the path is in the current_repo, and then returning something based on that.

tests/test_webeditorposts.py
```diff
@@ -256,16 +233,15 @@ def test06_test_publish_button(self):
...
+        rv = self.client.get('/delete_post?path={}'.format(self.post_id))
         assert rv.status == "200 OK"
 
         with self.app.app_context():
-            post = (db_session.query(Post)
-                              .filter(Post.id == self.post_id)
-                              .first())
-            assert not post
+            kp = self.repo.post(self.post_path)
+            assert kp.is_deleted is True
```
NiharikaRay: question - would `assert kp.is_deleted` be equivalent to `assert kp.is_deleted is True`

NiharikaRay:
Thanks for refactoring all of this Dan!

knowledge_repo/app/index.py
danfrankj: @matthewwardrop in python 3 if you try to compare None to an integer, it throws an exception

PR #105, opened 2016-11-02
Reindex in a background thread on demand.

matthewwardrop:
Currently, the index is only updated once immediately after being launched. Reindexing the repository requires restarting the app, which can take some time depending on the size of the repository. This patch changes this behaviour to reindex on demand. That is, every time a request is made, knowledge repo checks whether it should update the index. Knowledge repo will never update the index if another index is already underway, or if it is less than five minutes since the last index. It will otherwise re-index if the KnowledgeRepository(-ies) report a newer version than what is recorded in the index. When an index is being performed, it is done in a background thread. The status of the index is noted in the footer of the page, which indicated how long ago the index was performed (or if the index is currently being performed). An example is shown below:

screen shot 2016-11-02 at 2 19 48 pm

This PR would currently have some race conditions when using SQLite databases due to the frequent use of `db_session.flush()` in `_update_index`, and so background threading for the index update is disabled for sqlite databases.


knowledge_repo/app/app.py
```diff
@@ -96,33 +97,12 @@ def show_traceback(self):
...
             excluded_tags = current_app.config.get('EXCLUDED_TAGS', [])
             for tag in excluded_tags:
-                tag_exists = (db_session.query(Tag)
-                                        .filter(Tag.name == tag)
-                                        .all())
-                if not tag_exists:
-                    tag_exists = Tag(name=tag)
-                    db_session.add(tag_exists)
-                    db_session.commit()
+                Tag(name=tag)
```
NiharikaRay: [question] can you just do this since duplicate tags won't be added?

knowledge_repo/app/index.py
```diff
@@ -1,61 +1,117 @@
...
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
 
+def is_indexing():
+    return int(IndexMetadata.get('lock', 'index', '0'))
+
+def seconds_since_index():
+    last_update = IndexMetadata.get_last_update('lock', 'index')
+    if last_update is None:
```
NiharikaRay:
```
if last_update:
   return (datetime.datetime.utcnow() - last_update).total_seconds()
return None
```

knowledge_repo/app/index.py
```diff
@@ -1,61 +1,117 @@
...
+        return None
+    else:
+        return (datetime.datetime.utcnow() - last_update).total_seconds()
+
+def human_readable_time_since_index():
+    if is_indexing():
+        return 'Currently indexing'
+    seconds = seconds_since_index()
+    if seconds is None:
+        return "Never"
```
NiharikaRay: same None comment from above (None is inherently falsey)

knowledge_repo/app/templates/base.html
```diff
@@ -104,7 +104,8 @@
         </div>
 
         <div class="footer">
-            Served with ♥ by <a href="https://github.com/airbnb/knowledge-repo">Knowledge Repo</a> <a title='{{version_revision}}' href="https://github.com/airbnb/knowledge-repo/releases/tag/v{{ version }}">{{ version }}</a>
+            Served with ♥ by <a href="https://github.com/airbnb/knowledge-repo">Knowledge Repo</a> <a title='{{version_revision}}' href="https://github.com/airbnb/knowledge-repo/releases/tag/v{{ version }}">{{ version }}</a><br />
```
NiharikaRay: haha in theory we should make this an actual glyphicon

knowledge_repo/app/models.py
```diff
@@ -26,6 +26,39 @@
...
+    def get(cls, type, name, default=None):
+        m = db_session.query(IndexMetadata).filter(IndexMetadata.type == type).filter(IndexMetadata.name == name).first()
+        if m is not None:
+            return m.value
+        return default
+
+    @classmethod
+    def set(cls, type, name, value):
+        m = db_session.query(IndexMetadata).filter(IndexMetadata.type == type).filter(IndexMetadata.name == name).first()
+        if m is not None:
```
NiharikaRay: same None comment

knowledge_repo/app/models.py
```diff
@@ -26,6 +26,39 @@
...
+            m.value = value
+        else:
+            m = IndexMetadata(type=type, name=name, value=value)
+            db_session.add(m)
+
+    @classmethod
+    def get_last_update(cls, type, name):
+        m = db_session.query(IndexMetadata).filter(IndexMetadata.type == type).filter(IndexMetadata.name == name).first()
+        if m is not None:
```
NiharikaRay: same None comment

knowledge_repo/app/app.py
matthewwardrop: The way I crafted the model for `Tag` objects means that tags are deduplicated as a matter of course. There can never be duplicate tags when done this way.

knowledge_repo/app/models.py
```diff
@@ -26,6 +27,39 @@
 db_session = LocalProxy(lambda: current_app.db.session)
 
 
+class IndexMetadata(db.Model):
+    __tablename__ = 'index_metadata'
+
+    id = db.Column(db.Integer, nullable=False, primary_key=True)
+    type = db.Column(db.String(255), nullable=False)
```
danfrankj: are you imagining that type and name will be jointly unique?

danfrankj: if so, we should endcode this table that way

danfrankj: also do we need type and name?

knowledge_repo/app/index.py
```diff
@@ -1,17 +1,99 @@
...
+
+def update_index_required():
+    if not current_app.config.get('REPOSITORY_INDEXING_ENABLED', True):
+        return False
+
+    seconds = seconds_since_index()
+    seconds_check = seconds_since_index_check()
+    if is_indexing() or (seconds is not None) and (seconds < 5 * 60) and (seconds_check < 5 * 60):
+        return False
+    try:
```
danfrankj: nit: this reads a little strangely without an except block.

knowledge_repo/app/index.py
```diff
@@ -1,17 +1,99 @@
...
+        return "{:d} seconds ago".format(int(round(seconds)))
+    else:
+        return "{:d} minutes ago".format(int(round(seconds / 60)))
+
+
+def update_index_required():
+    if not current_app.config.get('REPOSITORY_INDEXING_ENABLED', True):
+        return False
+
+    seconds = seconds_since_index()
```
danfrankj: small efficiency gains if you bail early here

knowledge_repo/app/models.py
matthewwardrop: Yes; tuples of `type` and `name` should be unique. I'll encode this into the table as you suggest. We need type and name for things like:
type='repository_revision', name='<repository uri>', value='<revision>'

It would be possible to cram these into one field, but I felt like that was a bit clumsy. Agreed?

knowledge_repo/app/index.py
matthewwardrop: I think this is actually kosher Python code. It's the neatest way to guarantee that code block always runs even if there are exceptions (which are not caught).

knowledge_repo/app/models.py
danfrankj: yup sounds good

sryza:
@matthewwardrop when you say "so background threading for the index update is disabled for sqlite databases", does this mean that, when using the default database for the index, web server restarts are still required to pick up repository changes?

matthewwardrop:
Hi @sryza! No... It just means that when the git repository is updated, there will be a synchronous updating of the repository index. Depending on the size of the repository, this could be slow... But I strongly advise against using sqlite databases in production. For testing or for small servers it probably is not terrible.

sryza:
Great, thanks for the clarification!

PR #107, opened 2016-11-03
Fix postgres empty string bug

davidlundgren:
Creating the initial comment of a post when using postgres fails with the following
error:

```
DataError: (psycopg2.DataError) invalid input syntax for integer: ""
```


knowledge_repo/app/routes/comment.py
```diff
@@ -35,9 +35,13 @@ def post_comment():
     if not post:
         raise Exception('Unable to find post')
 
-    comment = (db_session.query(Comment)
-                         .filter(Comment.id == comment_id)
-                         .first())
+    if comment_id is not None:
+        comment = (db_session.query(Comment)
```
NiharikaRay: I believe this is equivalent to `if comment_id`, since `comment_id` will either be an integer or `None`

davidlundgren: Ah, I agree. I was being cautious for `comment_id`s with the value zero, but that probably doesn't happen!

Updated!

NiharikaRay:
You're all good to merge!

davidlundgren:
Sweet! I'm not sure if I can merge it, I'm seeing this message:

screen shot 2016-11-03 at 3 30 20 pm

NiharikaRay:
The builds have been a little slow recently :/ the way to make this faster is to work off of a fork, but for now we may just need to wait a little bit for the builds to catch up

NiharikaRay:
Looks like this is green! Feel free to merge whenever :)

davidlundgren:
Thanks @NiharikaRay, unfortunately it looks like I need write access to merge. Could you merge them?
screen shot 2016-11-03 at 8 33 21 pm

PR #108, opened 2016-11-03
[Permissions] Add parsing + front-end

NiharikaRay:

The long awaited PR that adds the front-end + parsing to groups. The best way to explain this is through a combination of words and pictures, so here goes. 

In the YAML headers, if there's an private flag, we add the appropriate status, and add the groups to the groups associated with the post. By default, we add a group, entitled `group.post_id`, that is the "custom group" per post. 

If you are the author of a post, and your post has the private flag, you will see this link at the top of your rendered post with "Add Permissions" as the text (this is where the old "Edit on Github" links were)

This brings you to this page
screen shot 2016-11-03 at 1 22 39 am

Clicking on "View Users" brings you to:
screen shot 2016-11-03 at 1 22 45 am

And on "Add Users", you get a typeahead, similar to what's on the Post Editor
screen shot 2016-11-03 at 1 22 57 am

If you don't have permission to view a post, you get this page:
screen shot 2016-11-03 at 1 25 38 am

PTAL. 

There are a few features missing, which I'll put in a todo list

- [x] Allow knowledge admins to have access to all posts (just extend the if statement)
- [x] Add a route where you can see all groups available (/all_groups), that also allows you to add a group
- [ ] Add Unit Tests

That being said, I'd love to get eyes on this ASAP since this was a feature request by users.

The reason there are 2 commits is to break up the work done - the first commit has the post processing parts, while the second commit is all the front-end work.

Once this is in, I'll tag a new release. 

cc: @akuhn

### DIFF
--- a/knowledge_repo/app/models.py
+++ b/knowledge_repo/app/models.py
@@ -350,7 +350,7 @@ class Post(db.Model):
 
         # create an implicit group, group_post.id, to add
         # single users to
-        group = Group(name="group_" + str(self.id))
+        group = Group(name=":post_group_" + str(self.id))
 
         # this created group should have the author associated with it
         # so they can add people to the post
@@ -502,6 +502,8 @@ class Post(db.Model):
 
         self.status = kp.status
 
+        self.private = 0
+        # we do this check so that no header (None) and False are treated the same
         if headers.get('private', ''):
             self.private = 1
             self.groups = headers.get('allowed_groups', [])
@@ -542,4 +544,4 @@ class Group(db.Model):
 
     @users.setter
     def users(self, user_objs):
-        self._users = user_objs
+        self._users = self._users + user_objs
--- a/knowledge_repo/app/models.py
+++ b/knowledge_repo/app/models.py
@@ -350,8 +350,13 @@ class Post(db.Model):
 
         # create an implicit group, group_post.id, to add
         # single users to
-        group = Group(name="group_" + self.id)
+        group = Group(name="group_" + str(self.id))
+
+        # this created group should have the author associated with it
+        # so they can add people to the post
+        group.users = self.authors
         group_objs.append(group)
+
         self._groups = group_objs
 
     _status = db.Column('status', db.Integer(), default=0)
@@ -498,8 +503,8 @@ class Post(db.Model):
         self.status = kp.status
 
         if headers.get('private', ''):
-            self.private = 0
-            self.groups = headers.get('allowed_groups')
+            self.private = 1
+            self.groups = headers.get('allowed_groups', [])
 
 
 class Email(db.Model):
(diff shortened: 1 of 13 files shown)


NiharikaRay:
Here's another screenshot, this time of the all_groups page, where you can add and view groups.

screen shot 2016-11-03 at 1 59 56 am

knowledge_repo/app/models.py
```diff
@@ -497,9 +507,11 @@ def update_metadata_from_kp(self, kp):
 
         self.status = kp.status
 
+        self.private = 0
+        # we do this check so that no header (None) and False are treated the same
         if headers.get('private', ''):
```
matthewwardrop: This one should be: `headers.get('private', False)`, since the types are translated to boolean if present. Obviously doesn't change logic, so not a biggie.

knowledge_repo/app/models.py
```diff
@@ -350,8 +350,18 @@ def groups(self, groups):
 
         # create an implicit group, group_post.id, to add
         # single users to
-        group = Group(name="group_" + self.id)
+        # we don't have self.id here necessarily, so query the Post table
```
matthewwardrop: Hi Nikki! Under which circumstances do we not have the id? If the id can be obtained using the query below, then it should also be available via self.id, no? What am I missing?

knowledge_repo/app/models.py
```diff
@@ -350,8 +350,18 @@ def groups(self, groups):
...
+        # for the path
+        post_id = (db_session.query(Post)
+                             .filter(Post.path == self.path)
+                             .first()).id
+        group = Group(name="group_" + str(post_id))
```
matthewwardrop: Do you think we should only create these implicit groups on demand? This creates the implicit post-specific group even if it will always be empty. Also, should we call the group something which cannot conflict with the set of other group names? Like... ':post_<id>', and then prevent the colon character from being used otherwise? I suggest it now only so we don't need to do migrations later if we do decided to do it cleanly in this way.

matthewwardrop:
Hi Nikki. This is looking good. I focussed more on the data model in my review than in the interface, since the data model is harder to fix in subsequent prs (due to migration of live data). My main suggestion suggestion is to name the default groups something like ':post_<id>', and then we can later lock it down so other groups cannot begin with a colon (or some such). It also allows us to add other implicity/virtual groups as necessary later on. Other than that, my comments were pretty minor.

With this fix, I'm happy for the pr to be merged as is; and we can fix any other details as necessary later.

LGTM.

knowledge_repo/app/models.py
NiharikaRay: I tested this in two ways, the first was by deleting my original db so that all posts would be re-added. In that case, there was no `self.id`, which was interesting to me since we actually add the post. From this post: http://stackoverflow.com/questions/23076483/sqlalchemy-get-added-objects-id-without-committing-session, it looks like we may need to flush the object, so I can try that to see if we can just use the id, since that's far less brittle

knowledge_repo/app/models.py
NiharikaRay: I think not, I think it's useful to always just have a default group available for people to add to. But I'll definitely rename it so it doesn't conflict with any user specified group names

PR #110, opened 2016-11-03
Fix bytes encoding bug between Python 2 and 3

davidlundgren:
In Python 2, `str.encode()` returns an encoded `str`ing object, while in Python
3, the same call returns a `bytes` object. This mismatch causes display issues
on the "View Raw Markdown" page. (H/T @b11z)

On `master`:
image

Post-PR:
screen shot 2016-11-03 at 2 23 35 pm


NiharikaRay:
Feel free to merge this! all checks are green!

PR #112, opened 2016-11-03
[Github] Auto-tag people in issues + PRs

NiharikaRay:
As contributions grow, we should probably have all of us tagged in PRs to increase throughput and on issues, to increase visibility on them.

@NiharikaRay @matthewwardrop @earthmancash @danfrankj

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,2 @@
+
+Auto-reviewers: @NiharikaRay @matthewwardrop @earthmancash @danfrankj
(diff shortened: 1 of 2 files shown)


matthewwardrop:
Sounds good. I wonder if we should use a checklist as in knowledge repo. In most cases, I don't know what I'd put under "Test Plan", so I'd probably delete it.

NiharikaRay:
That's fair, I think it's good practice to at least mention in the PR how it was tested, so if someone is more familiar with certain parts of the code, they can make sure all relevant tests are done. Regardless, we can change this after the fact, I mostly want this so we're all tagged on all PRs.

PR #114, opened 2016-11-04
Add downgrade actions to dbs

NiharikaRay:
**Description of changeset:** 
In addition to having an upgrade function for databases, it is nice to have a downgrade one (which I found was necessary when testing the permission changes)

**Test Plan:** 
Run the command, both with supplying a revision number and not, and expect it to run - it fails on sqlite databases due to the lack of alter column support

Also, made a slight change to the PR template, after dogfooding it
Auto-reviewers:  @matthewwardrop @earthmancash @danfrankj


NiharikaRay:
Sure, I can fix that!

Issue #65, opened 2016-10-12
knowledge_repo script review


matthewwardrop:
The `knowledge_repo` script has grown organically since it was created, and there are a few confusing properties of the script (such as when to use the `-p` option with the `add` action). It will be thoroughly reviewed, and updated accordingly.

PR #121, opened 2016-11-06
Use pallets/click for argument parsing and command dispatching

FooBarQuaxx:
Description of changeset: 
- Use click as the command line commands and argument handler.

Test Plan: 
- Insure that the behavior of the application does not regress.

### DIFF
--- a/knowledge_repo/_version.py
+++ b/knowledge_repo/_version.py
@@ -34,6 +34,7 @@ __dependencies__ = [
     'werkzeug',  # Development webserver
     'gunicorn',  # Deployed webserver
     'inflection',  # String transformation library
+    'click',  # argument parsing and command dispatching
 ]
 
 __optional_dependencies__ = {
--- a/knowledge_repo/repositories/gitrepository.py
+++ b/knowledge_repo/repositories/gitrepository.py
@@ -76,7 +76,7 @@ class GitKnowledgeRepository(KnowledgeRepository):
 
     @path.setter
     def path(self, path):
-        assert isinstance(path, str), "The path specified must be a string."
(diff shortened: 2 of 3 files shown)


matthewwardrop:
Hi @FooBarQuaxx,

Thanks for taking the time to work on improving the Knowledge Repository. We really appreciate it. It is with help from people such as yourself that we have a chance to make it something really great.

I scanned through the patch quickly, and noted at least one regression. In particular, it loses the ability to use a nested instance of a knowledge repository (`knowledge_repo` is already imported, and so changing the path to the location of the new repository is not going to help in `main()`). Other than that it looks really promising and quite a bit nicer than the standard `argparse` approach. We previously avoided such an approach so as to stay close to mainline Python in this regard.

Nevertheless, for now, I'm not going to approve it for merging into master. The reason for this is simply that we're about to review the knowledge repo script anyway; and going through fixing these regressions and checking everything with a fine comb might as well be rolled into that process.

With your permission, I might base some of that rebasing work on top of this patch-set, if we decide to adopt `click`. Do I have such permission? If we do you use it, I'll keep your name in the commit log.

For now, I'll close this PR; and pick it up again in #65 .

Thanks again!

FooBarQuaxx:
If it's possible to fix that regression ( which is not covered by the tests) I would like to do it. Otherwise no problem if you use my work in further work.

matthewwardrop:
It should be easy to fix by simply importing knowledge_repo after changing the path in main(), rather than at the outset of the file. I haven't closely checked for other regressions, but using click makes it easier to see such things, and so I do see a high likelihood of us using click going forward.

Again, really appreciate your efforts. I'll tag you on future work regarding this.

PR #122, opened 2016-11-06
Minor CLI improvements for help, exit codes

parente:
* Let --help/-h fall through to where subcommands are defined so the
  list of them can be seen
* Show abbreviated top-parser help when --repo is not set to retain a
  hint of what the user needs to do
* Fix formatting of the ValueError
* Fix odd git interrupted errors, missing subprocess return code by
  waiting for the subprocess to complete
* Fix case where no args are specified at all, show the help

### DIFF
--- a/scripts/knowledge_repo
+++ b/scripts/knowledge_repo
@@ -82,8 +82,8 @@ if args.version:
 
 if args.repo is None:
     parser.print_help()
-    raise ValueError("No repository specified. Please set the --repo flag, \
-or the KNOWLEDGE_REPO environment variable.")
+    raise ValueError("No repository specified. Please set the --repo flag, "
+                     "or the KNOWLEDGE_REPO environment variable.")
 
 # Update repository so that we can ensure git repository configuration is up to date
 # We wrap this in a try/except block because failing to update a repository can
--- a/scripts/knowledge_repo
+++ b/scripts/knowledge_repo
@@ -80,13 +80,10 @@ args, remaining_args = parser.parse_known_args()
 if args.version:
     print('Local version: {}'.format(knowledge_repo.__version__))
 
-if args.help and not remaining_args:
-    parser.print_help()
-    sys.exit(0)
-
 if args.repo is None:
+    parser.print_help()
     raise ValueError("No repository specified. Please set the --repo flag, \
-                      or the KNOWLEDGE_REPO environment variable.")
+or the KNOWLEDGE_REPO environment variable.")
 
 # Update repository so that we can ensure git repository configuration is up to date
 # We wrap this in a try/except block because failing to update a repository can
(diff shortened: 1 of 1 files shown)


scripts/knowledge_repo
```diff
@@ -80,13 +80,10 @@ args, remaining_args = parser.parse_known_args()
...
 
-if args.help and not remaining_args:
-    parser.print_help()
-    sys.exit(0)
-
 if args.repo is None:
+    parser.print_help()
     raise ValueError("No repository specified. Please set the --repo flag, \
-                      or the KNOWLEDGE_REPO environment variable.")
+or the KNOWLEDGE_REPO environment variable.")
```
NiharikaRay: We've settled on aligning everything with the opening delimiter (https://www.python.org/dev/peps/pep-0008/#indentation) - there are more examples of this in the actual app code.

scripts/knowledge_repo
```diff
@@ -110,8 +107,9 @@ if (isinstance(args.repo, str) and not args.dev
         and os.path.exists(os.path.join(args.repo, '.resources', 'scripts', 'knowledge_repo'))
         and os.path.abspath(args.repo) != os.path.abspath(os.path.join(os.path.dirname(__file__), '../../'))):
     cmdline_args = ['--noupdate'] + [arg.replace(' ', '\ ') if ' ' in arg else arg for arg in sys.argv[1:]]
-    subprocess.call('{} {}'.format(os.path.join(os.path.abspath(args.repo), '.resources/scripts/knowledge_repo'), ' '.join(cmdline_args)), shell=True)
-    sys.exit(0)
+    proc = subprocess.call('{} {}'.format(os.path.join(os.path.abspath(args.repo), '.resources/scripts/knowledge_repo'), ' '.join(cmdline_args)), shell=True)
+    return_code = proc.wait()
+    sys.exit(return_code)
```
NiharikaRay: @matthewwardrop you are more familiar with the CLI code so I'll defer to you on some of this.

scripts/knowledge_repo
parente: The problem here is that when the exception appears in the CLI, there is a ton of whitespace in the string.

matthewwardrop:
Hey @parente ,

Thanks again for taking the time to work on the knowledge repository. Most of the behaviour which you are "fixing" was actually an intentional decision; but that may simply mean that the script is broken by design.

The problem is that we don't guarantee future versions of the script have exactly the same API (we are not API stable yet), so we wait until we know which version of the script is going to be run before showing the possible actions (rather than potentially showing the wrong thing). Of course, in most cases, the API between the scripts is backward compatible (as in, the delegated version of the script has a superset of actions compared to the pip installed version); but this is not necessarily the case.

The use case we are serving here is when a user is contributes to multiple different knowledge repositories, each running a different version on the server-side. Also, in large organisations, it is handy to upgrade everyone simultaneously, so we have this 'delegated' way to launch the knowledge_repo script appropriate for each repository; rather than have people pip upgrade in dribs and drabs (which means that it is easier to roll out bugfixes).

It is unclear to me whether we should adopt your changes here (and hence sometimes be wrong), or just signpost to the user that to see the help they must provide enough information to identify the version of the script to be run.

What are your thoughts on this?

P.S. We are going to be revamping the knowledge_repo script soon, but I expect these issues to remain relevant.

scripts/knowledge_repo
parente: Can do!

parente:
> The problem is that we don't guarantee future versions of the script have exactly the same API (we are not API stable yet), so we wait until we know which version of the script is going to be run before showing the possible actions (rather than potentially showing the wrong thing). 

I figured this was the case given that the the script runs a possibly older or newer copy of itself as a subprocess part way through. I thought I retained the correct behavior for this case, however, since the subprocess call comes before adding any of the subparsers. But I'll admit I don't have a perfect setup to test this case.

> It is unclear to me whether we should adopt your changes here (and hence sometimes be wrong), or just signpost to the user that to see the help they must provide enough information to identify the version of the script to be run.

Your choice. Again, I was aiming to retain the delegation model, and may have missed the mark.

matthewwardrop:
I looked over this more carefully, and it looks good. Merging into master now.

PR #125, opened 2016-11-07
Enable DBRepository to work with Postgres and fix a couple errors

sryza:
This fixes a couple issues I ran into when trying to run the knowledge repo over a postgres DB.  Without these changes, I wasn't able to get commands like `knowledge_repo add` to complete, and with them, I was.

### DIFF
--- a/knowledge_repo/repositories/dbrepository.py
+++ b/knowledge_repo/repositories/dbrepository.py
@@ -7,7 +7,7 @@ from sqlalchemy import create_engine
 from sqlalchemy.orm import sessionmaker, scoped_session, mapper
 from sqlalchemy.sql import func
 
-from sqlalchemy import Table, Column, Integer, String, DateTime, BLOB, MetaData
+from sqlalchemy import Table, Column, Integer, String, DateTime, LargeBinary, MetaData
 from ..repository import KnowledgeRepository
 
 logger = logging.getLogger(__name__)
@@ -44,7 +44,7 @@ class DbKnowledgeRepository(KnowledgeRepository):
                               Column('revision', Integer, default=0),
                               Column('status', Integer, default=self.PostStatus.DRAFT.value),
                               Column('ref', String(512)),
-                              Column('data', BLOB))
+                              Column('data', LargeBinary))
         self.engine = create_engine(engine_uri)
         self.session = scoped_session(sessionmaker(bind=self.engine))
         if auto_create:


knowledge_repo/repositories/dbrepository.py
```diff
@@ -99,10 +99,10 @@ def _dir(self, prefix, statuses):
 
     # -------------- Post submission / addition user flow --------------------
 
-    def _add_prepare(self, kp, path, update=False):
+    def _add_prepare(self, kp, path, update=False, branch=None, squash=None, message=None):
         pass
 
-    def _add_cleanup(self, kp, path, update=False):
+    def _add_cleanup(self, kp, path, update=False, branch=None, squash=None, message=None):
```
earthmancash: What do you need these optional fields for?

earthmancash:
Hey Sandy, thanks for the contribution! I'm doing some testing to see if switching from BLOB to LargeBinary on our MySQL setup will require a migration, in which case we may want to think more internally about how to stage it. Left one comment about the optional subfields that you added though, what do you need them for? 

cc @NiharikaRay

sryza:
@earthmancash I don't need those fields, but the code that calls those functions passes those fields in through kwargs.  I added the fields because the function invocations were failing without them.   

LargeBinary is a SQLAlchemy type that translates to BLOB on MySQL:
http://docs.sqlalchemy.org/en/latest/core/type_basics.html#sqlalchemy.types.LargeBinary

earthmancash:
Hrmm, there aren't really the concepts of branches/squashing on dbrepo posts, so I'd prefer not to include those fields. I'm not seeing the function invocation failures, can you paste the output? 

The LargeBinary changes makes sense to me, thanks for the contribution!

sryza:
Here's the error, with Python 3.5:

```
Traceback (most recent call last):
  File "/Users/sandyryza/.pyenv/versions/3.5.1/bin/knowledge_repo", line 6, in <module>
    exec(compile(open(__file__).read(), __file__, 'exec'))
  File "/Users/sandyryza/knowledge-repo/scripts/knowledge_repo", line 249, in <module>
    repo.add(kp, path=args.path, update=args.update, branch=args.branch, message=args.message, squash=args.squash)
  File "/Users/sandyryza/knowledge-repo/scripts/../knowledge_repo/repository.py", line 226, in add
    cleanup_kwargs = self._add_prepare(kp, path, update, **kwargs)
TypeError: _add_prepare() got an unexpected keyword argument 'squash'
```

I'm relatively new to Python, but in most programming languages I've interacted with, subclass methods that override a superclass method need to accept all the arguments that the superclass method accepts, even if they don't use them for anything.  Otherwise, the caller can't be agnostic to the implementation behind the interface.

earthmancash:
Ok, I think I know the issue. We use the dbrepo only through the metarepo over here, which handles those invocations better. We're going to have a discussion about the best way to support knowledge app instances with just a dbrepo, which will likely be something like having a metarepo around single repo specified webapps.

For now, this LGTM

matthewwardrop:
@sryza @earthmancash 

Thanks again for your patch @sryza; but if possible, can you please replace the extra fields on _add_prepare and _add_cleanup with `**kwargs`? These fields are currently being passed in from the knowledge_repo script, and there is no explicit (or implicit) need for them to be used in the `DBKnowledgeRepository` object. Once you've done this, we'll gladly merge it!

@earthmancash Wrapping it in a `MetaKnowledgeRepository` won't help either, since it will pass through options from the `knowledge_repo` script; which is currently only designed for git repositories. Fixes for that incoming.

sryza:
Updated the PR

matthewwardrop:
Excellent. Thank you! Merging now.

PR #130, opened 2016-11-08
bugfix to allow reindexing

danfrankj:
@matthewwardrop 

Previously current_repo.revisions returned a mapping from mountpoint to Repository object which can't be queried

### DIFF
--- a/knowledge_repo/repositories/dbrepository.py
+++ b/knowledge_repo/repositories/dbrepository.py
@@ -64,7 +64,7 @@ class DbKnowledgeRepository(KnowledgeRepository):
 
     @property
     def revision(self):
-        return self.session.query(func.max(self.PostRef.created_at))
+        return self.session.query(func.max(self.PostRef.created_at)).first()[0]
 
     def update(self):
         pass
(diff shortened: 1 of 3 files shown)


knowledge_repo/repositories/dbrepository.py
```diff
@@ -64,7 +64,7 @@ def session_end(self):
 
     @property
     def revision(self):
-        return self.session.query(func.max(self.PostRef.created_at))
+        return self.session.query(func.max(self.PostRef.created_at)).first()[0]
```
NiharikaRay: is it possible that there is nothing returned, so the `.first()[0]` will fail?

NiharikaRay:
also, I'm sure you'll fix the builds before merging :)

knowledge_repo/repositories/dbrepository.py
danfrankj: Even when the table is empty this query should return (None,)

matthewwardrop:
I quickly fixed up the commit as discussed with @danfrankj. Thanks for noticing it though... I skimped when checking this one ><.

Issue #115, opened 2016-11-04
Adding a long commit message


kaelhuerta:
Auto-reviewers: @NiharikaRay @matthewwardrop @earthmancash @danfrankj

Hey, reviewers! I'm gathering several questions that I had while testing the repo out. Please let me know if you prefer an issue for each of them.

1. Is there a way of adding a long message to the commits instead of the default `commit -m` style?

2. Why are so many pages displayed even if I have only 2 posts?

![image](https://cloud.githubusercontent.com/assets/6402208/20034810/5bed7006-a391-11e6-977c-8c4a39df9779.png)

3. How do I change the logo? Actually, is not appearing in the server, but when I run the exact same thing locally it does show (the `/static/images/logo-white.svg` one)

4. What are your suggestions in order to change the complete blog theme? Fork this repo and change it?, can it be done with a local config file that I have missed so I always point to the main repository instead of a forked one?

5. My posts are all showing the default thumbnail, how to select which graph to show?

6. Can a button be added (or the one that is there already changed) to toggle the `.Rmd` file instead of the `.md`? This could address #118 in a more elegant way. You write your post hiding all the code for the non-technical readers but allow the curious one to see how it was created by toggling to the raw `.Rmd`

Thanks for the hard work, I'm really enjoying testing it out.
Kael

matthewwardrop:
Hi @kaelhuerta,

Thanks for your interest in the knowledge repository :). Let me respond to your questions one by one :).

> Is there a way of adding a long message to the commits instead of the default commit -m style?
At this stage, no. We will be renovating the `GitKnowledgeRepository` class soon, so we'll try to add such support then.

> Why are so many pages displayed even if I have only 2 posts?
The user interface is quite immature, since a lot of recent attention has been spent on improving the back end. The user interface will get incrementally improved over time.

> How do I change the logo? Actually, is not appearing in the server, but when I run the exact same thing locally it does show (the /static/images/logo-white.svg one)
> What are your suggestions in order to change the complete blog theme? Fork this repo and change it?, can it be done with a local config file that I have missed so I always point to the main repository instead of a forked one?
[...]

Issue #51, opened 2016-10-06
Add a custom thumbnail in the feed page


qthurier:
Is there a way to do that during the adding step of a knowledge post? The is a feed_image parameter in the code but I don't see how to use it.

matthewwardrop:
Hey @qthurier! We do have a patch against the old internal version of this code which adds support for just that. We're almost finished deprecating our internal repository, after which time all development will be done here. I'll see if I cannot resurrect it soon.

The way this feature currently is implemented is to allow you to specify in the headers of your post which image should be used as the thumbnail of the post. If none is provided, it will use the first image in the post (as currently). Would that work for you?

qthurier:
Yeah that would be great thanks!

matthewwardrop:
Support now complete in master.

qthurier:
Humm could you provide an example of what the header should like then? 

This doesn't work (I still have the default image):

---
title: An unsuccefull attempt to ..
authors:
- Quentin
tags:
- tag0
- tag1
thumbnail: 2
created_at: 2016-11-11
updated_at: 2016-11-11
tldr: This notebook is extracted from the bitbucket repo ...
---

matthewwardrop:
Hi @qthurier,

You should run this header through add script, which will then hard code the actual path into the the thumbnail field. I've tested it locally, and it all seems to work well. Can you run through how you are adding it to the repository?

Also, can you run `knowledge_repo --repo <repo> version`?

qthurier:
I added the header as a raw nbconvert cell at the top of my notebook, then this sequence of actions: add, submit (merge the new branch) and deploy

knowledge_repo --repo ds_knowledge_repo/ --version :
```
Local version: 0.7.2
INFO:knowledge_repo.repositories.gitrepository:Fetching updates to the knowledge repository...
usage: knowledge_repo [--repo REPO] [--knowledge-branch KNOWLEDGE_BRANCH]
                      [--dev] [--debug] [--noupdate] [-h]
                      {init,create,drafts,status,add,submit,push,preview,runserver,deploy,db_upgrade}
                      ...
knowledge_repo: error: unrecognized arguments: --version
```

PR #138, opened 2016-11-10
Add support for specifying a thumbnail image, and fix some bugs in the post caching mechanism.

matthewwardrop:
This addresses #51 and the issues in the discussion of #115.

Thumbnails can be specified as integers (in which case the nth image is chosen as the thumbnail) or strings (in which case the string is treated as a url). The resulting url is then hard coded as the thumbnail header of the generated markdown file, which means that if this logic changes in the future, old posts will not be affected.

### DIFF
--- a/knowledge_repo/postprocessors/extract_images.py
+++ b/knowledge_repo/postprocessors/extract_images.py
@@ -13,8 +13,38 @@ class ExtractImages(KnowledgePostProcessor):
     @classmethod
     def process(cls, kp):
         images = cls.find_images(kp.read())
-        cls.collect_images(kp, images)
+        image_mapping = cls.collect_images(kp, images)
+        cls.update_thumbnail_uri(kp, images, image_mapping)
         cls.cleanup(kp)
+
+    @classmethod
+    def update_thumbnail_uri(cls, kp, images, image_mapping):
+        thumbnail = kp.headers.get('thumbnail', 0)
+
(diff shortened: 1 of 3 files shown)


earthmancash:
This LGTM. Might be worth adding it to the template files, even manually setting the image index to 0 in them, to make the feature clear.

knowledge_repo/postprocessors/extract_images.py
```diff
@@ -13,10 +13,37 @@ class ExtractImages(KnowledgePostProcessor):
...
+        image_mapping = cls.collect_images(kp, images)
+        cls.update_thumbnail_uri(kp, images, image_mapping)
         cls.cleanup(kp)
 
     @classmethod
+    def update_thumbnail_uri(cls, kp, images, image_mapping):
+        thumbnail = kp.headers.get('thumbnail', 0)
+
+        # if thumbnail is a number, select the nth image in this post as the thumbnail
+        if isinstance(thumbnail, str) and thumbnail.isdigit():
```
mistercrunch: use `basestr` if you want unicode to work in py27

matthewwardrop: Good catch @mistercrunch .

matthewwardrop: I'll handle this in a subsequent patch, since this is also an issue in several other places.

PR #140, opened 2016-11-10
a quick fix for handling of prefixes that are web-editable

danfrankj:
WIP

Do we only want to allow webeditor if the WEB_EDITOR_PREFIXES has a list of prefixes? 

### DIFF
--- a/knowledge_repo/app/routes/web_editor.py
+++ b/knowledge_repo/app/routes/web_editor.py
@@ -1,13 +1,11 @@
-from builtins import range
-from builtins import str
 import json
 import logging
 import sys
 import os
-from collections import defaultdict
+from builtins import str
 from datetime import datetime
 from flask import request, render_template, Blueprint, current_app, url_for, send_from_directory, g
-from sqlalchemy import or_, distinct
+from sqlalchemy import or_
 from werkzeug import secure_filename
 
 from knowledge_repo.post import KnowledgePost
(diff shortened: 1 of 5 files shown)


matthewwardrop:
@danfrankj Perhaps we should allow editing everywhere if WEB_EDITOR_PREFIXES is None, but set the default configuration to be an empty list? At least until we get the rest of the backends up to speed?

knowledge_repo/app/routes/web_editor.py
```diff
@@ -1,13 +1,11 @@
-from builtins import range
-from builtins import str
 import json
 import logging
 import sys
 import os
-from collections import defaultdict
+from builtins import str
```
matthewwardrop: This is not a good idea, since it seems to break string handling in places outside of our project (i.e. not all packages seem to support the unicode strings that result...). That's been my experience anyway.

knowledge_repo/config_defaults.py
```diff
@@ -53,6 +53,10 @@ def web_uri(path=None):
     return None
 
 
+# Prefixes of repositories that can be edited via the web editor UI
+# Defaults to no prefixes allowed. If None, all prefixes editable via the UI.
+web_editor_prefixes = []
```
matthewwardrop: Shouldn't this key be specified in the server configuration, not the repository configuration? If not, then we need to propagate this configuration into methods on the `KnowledgeRepository` class and delegate in the `MetaKnowledgeRepository` class.

knowledge_repo/app/routes/web_editor.py
```diff
@@ -123,12 +124,23 @@ def save_post():
     data = request.get_json()
     path = data['path']
 
+    prefixes = current_app.config['web_editor_prefixes']
+    if prefixes == []:
+        raise Exception("Web editing is not configured")
```
matthewwardrop: Should this raise an exception?

PR #142, opened 2016-11-14
Cast port to int

mpekalski:
Description of changeset: 
Socket requires port to be int, added explicit cast to int and try/catch block to make sure it has been casted properly.

### DIFF
--- a/knowledge_repo/app/utils/render.py
+++ b/knowledge_repo/app/utils/render.py
@@ -77,7 +77,7 @@ def render_post(post):
     """
     def intra_knowledge_urlmapper(name, url):
         if name == 'a' and url.startswith('knowledge:'):
-            return url_for('render.render', markdown=url.split('knowledge:')[1])
+            return url_for('render.render', markdown=url.split('knowledge:')[1]).replace('%2F', '/')  # Temporary fix before url revamp
         return None
 
     html = render_post_header(post) + (post if isinstance(post, KnowledgePost) else post.kp).to_string('html',


NiharikaRay:
This looks good to me, thank you for your contribution!

knowledge_repo/repositories/gitrepository.py
```diff
@@ -503,15 +503,28 @@ def __remote_host(self):
...
 
     @property
+    def __remote_port(self):
+        port = 22
+        if self.git_has_remote:
+            m = re.match(r'^(.*?)?@([^/:]*):?(\d+)?', self.git_remote.url)
+            if m:
+                if m.group(3):
+                    port = m.group(3)
+        try:
```
matthewwardrop: You shouldn't need this try block, since the regex filter already restricts it to digits. Given that `\d` actually matches non western digit glyphs as well, it might be worth change it to: `[0-9]+` instead of `\d+`.

matthewwardrop:
Some of these changes have already made it into master; so please rebase on master :). After that, we'll gladly merge your changes in.

mpekalski:
Sorry I am not that proficient with git, I hope that is what you meant by rebasing.

NiharikaRay:
Hmm it looks like it wasn't rebased since there are still some conflicts. 

If you click on the "Use the command line" link above, you should be able to see the commands to correctly resolve the conflicts.

mpekalski:
Closed it by accident, but maybe for better. The new one looks ok now.
https://github.com/airbnb/knowledge-repo/pull/144

PR #144, opened 2016-11-15
changed \d to [0-9] and cast port to int

mpekalski:
Description of changeset: 
Port has to be explicitly casted to int , and [0-9] should be more strict than \d
Test Plan: 


matthewwardrop:
Looks great! Thanks @mpekalski!

PR #146, opened 2016-11-16
Small cleanups to indexing, along with some extra debugging tools.

matthewwardrop:
This patch set fixes a few small reindexing bugs, and adds some debug routes to assist in future problems like this one. In particular it addresses the following:
 - If a reindex has taken longer than 10 minutes, new reindexes are now allowed to occur; preventing a failed reindex / crashed server from preventing the index from being up to date.
 - A manual reindex can now be forced using the `reindex` action in the `knowledge_repo` script or (when running in debug mode) the `/debug/force_reindex` route.
 - Fix a small bug when chaining script and wanting to return the nested script's error code.
 - Fix S3 upload sample script for Python 3.
 - Add a debug route `/debug/versions` that shows the versions of both system and installed packages that are pertinent to the knowledge repository (see screenshot below).

screen shot 2016-11-16 at 1 00 23 pm

…

### DIFF
--- a/knowledge_repo/app/index.py
+++ b/knowledge_repo/app/index.py
@@ -15,27 +15,46 @@ logger = logging.getLogger(__name__)
 
 
 def is_indexing():
-    return int(IndexMetadata.get('lock', 'index', '0'))
+    timeout = current_app.config.get("INDEXING_TIMEOUT", 10 * 60)  # Default index timeout to 10 minutes (after which indexing will be permitted to run again)
+    last_update = time_since_index()
+    return bool(int(IndexMetadata.get('lock', 'index', '0'))) and last_update is not None and (last_update < timeout)
 
 
 def time_since_index(human_readable=False):
     last_update = IndexMetadata.get_last_update('lock', 'index')
-    return time_since(last_update, human_readable=human_readable)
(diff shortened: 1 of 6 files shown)


knowledge_repo/app/index.py
```diff
@@ -15,19 +15,36 @@
 
 
 def is_indexing():
-    return int(IndexMetadata.get('lock', 'index', '0'))
+    last_update = time_since_index()
+    return bool(int(IndexMetadata.get('lock', 'index', '0'))) and last_update is not None and (last_update < 10 * 60)
```
earthmancash: I wonder if we want to expose this 10 minute limit a bit more to make things more clear. Can put a max_index_time variable either in this function or server config. Without documenting the whole thing, maybe a docstring for the `_update_index` function would be helpful

earthmancash:
One nit on making the indexing logic a bit more clear, otherwise LGTM

knowledge_repo/app/index.py
matthewwardrop: Good call. I want to actually make this configurable (which is perhaps why I was lazy about documentation here)... but if I don't get to do it here, I'll document it as you proposed :).

PR #147, opened 2016-11-16
BadLock: Fix index lock check for postgres

davidlundgren:
This PR fixes the following error when running `knowledge-repo` with postgres:

```
    return int(IndexMetadata.get('lock', 'index', '0'))
ValueError: invalid literal for int() with base 10: 'false'
```

Test Plan: 
```bash
$ knowledge_repo --repo <repo> runserver --debug --dburi <some_postgres_db>
```

### DIFF
--- a/knowledge_repo/app/index.py
+++ b/knowledge_repo/app/index.py
@@ -15,7 +15,11 @@ logger = logging.getLogger(__name__)
 
 
 def is_indexing():
-    return int(IndexMetadata.get('lock', 'index', '0'))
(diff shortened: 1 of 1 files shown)


matthewwardrop:
Hi @davidlundgren; thanks for keeping us honest re-generic database support!

This change looks fine, but there may be other databases which also map the value differently. I think what we'll do instead is just be more specific about the string that gets entered into the database; that way it should work everywhere. I'll merge this into #146 .

Thanks again! Keep the fixes rolling in!

davidlundgren:
Thanks! Happy to help!

PR #148, opened 2016-11-18
add a reasonable height to proxy posts as well as a link to the original source

danfrankj:
Add a reasonable height to proxy posts and a link to their original source

### DIFF
--- a/knowledge_repo/converters/html.py
+++ b/knowledge_repo/converters/html.py
@@ -92,7 +92,7 @@ class HTMLConverter(KnowledgePostConverter):
 
         # proxy posts are assumed to be embeddable links
         if 'proxy' in self.kp.headers:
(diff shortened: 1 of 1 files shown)


matthewwardrop:
Perhaps we should adopt a solution like this: http://stackoverflow.com/questions/9162933/make-iframe-height-dynamic-based-on-content-inside-jquery-javascript

danfrankj:
@matthewwardrop - That solution doesn't work if the content is on a cross-domain (which is the idea here). It'd be great to have it set dynamically, but I'm not sure how to get around that issue.

matthewwardrop:
@danfrankj Good point. There doesn't seem to be a good general purpose solution to this that I can see (save perhaps proxying the page). But for now, this looks good to me.

Issue #94, opened 2016-10-28
Params do no get passed correctly to pagination


NiharikaRay:
If you click on an author in a knowledge post, you arrive at a feed for the author. Clicking the next arrow returns you to the second page of the feed, without the author being set. 

This is a bigger issue with how to deal with parameters (currently done hackily), I'll take this fix on.

earthmancash:
Fixed here https://github.com/airbnb/knowledge-repo/pull/150

PR #150, opened 2016-11-22
[Search] Fix parameter passing + author search

NiharikaRay:
Description of changeset: There was an issue (https://github.com/airbnb/knowledge-repo/issues/94) where certain parameters weren't being passed through correctly. I use javascript now to get existing parameters, and accordingly create the new hrefs. 

Also, cleaned up the way author search was done!

Test Plan: Ensure pagination works as expected on the /feed page

### DIFF
--- a/knowledge_repo/app/models.py
+++ b/knowledge_repo/app/models.py
@@ -310,7 +310,7 @@ class Post(db.Model):
     _authors = association_proxy('_authors_assoc', 'author',
                                  creator=lambda author: PostAuthorAssoc(author=author),)
 
-    @property
+    @hybrid_property
     def authors(self):
         return self._authors
 
--- a/knowledge_repo/app/templates/index-base.html
+++ b/knowledge_repo/app/templates/index-base.html
@@ -39,48 +39,31 @@
   <div class='col-md-3 col-md-offset-4'>
     {% set start = feed_params['start'] | default(0) %}
     {% set results = feed_params['results'] | default(10) %}
-    {% if start == 0 %}
-      {% set page = 1 %}
-    {% else %}
-      {% set page = start//results + 1 %}
-    {% endif %}
+    {% set page = 1 if (start == 0) else (start//results + 1) %}
+
     <div class="btn-group btn-sm btn-group-justified pagination-group" role="group">
-      {% if page == 1 %}
-        <a href="#" class="btn btn-default disabled">
+      {% set href = "#" if (page == 1) else request.path %}
+        <a href="{{ href }}" data-page="{{ page }}"
+           class="btn btn-default {{ "disabled" if (page == 1)}} prev-btn">
           <i class="glyphicon glyphicon-chevron-left"></i>
         </a>
-      {% else %}
-        {% set prev_page = url_for(request.endpoint, start=(page-2)*results, results=results, filters=feed_params['filters']) %}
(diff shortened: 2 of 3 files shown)


knowledge_repo/app/templates/index-base.html
```diff
@@ -98,12 +82,51 @@
 <script type="text/javascript">
 
     $( document ).ready(function() {
-     $.ajaxSetup({
-      cache: true
-    });
+      $.ajaxSetup({
```
danfrankj: for my education ... what does ajaxSetup do?

knowledge_repo/app/templates/index-base.html
```diff
@@ -98,12 +82,51 @@
...
+        cache: true
+      });
+
+      function getParameters() {
```
danfrankj: what about using the feed_params variable we have access to here?

danfrankj: We know what it'll have and we already do the parsing. We could potentially pass it into the function here as a string too to make it easy to tack onto the buttons

knowledge_repo/app/templates/index-base.html
```diff
@@ -39,48 +39,32 @@
...
         {% set page_nums = [page, page + 1, page + 2] %}
       {% else %}
         {% set page_nums = [page - 1, page, page + 1] %}
       {% endif %}
       {% for page_num in page_nums %}
-        {% set href = url_for(request.endpoint, start=(page_num - 1) * results, results=results, filters=feed_params["filters"]) %}
-          {% if request.endpoint == 'tag.render_tag_pages' %}
-            {% set href = href + "&tag=" + full_tag %}
-          {% endif %}
+        {% set href = request.path %}
```
danfrankj: why not just use request.path below

danfrankj: also it seems like we can us feed_params while we're setting these up rather than parsing them later

knowledge_repo/app/models.py
```diff
@@ -456,6 +456,15 @@ def view_user_count(self):
                 .where(PageView.object_type == 'post')
                 .label("view_user_count"))
 
+    @hybrid_property
+    def contains_author(self, author_name):
```
danfrankj: I must be missing something but I don't see this used anywhere.

knowledge_repo/app/templates/index-base.html
NiharikaRay: good call, I can use that below. The reason I can't use feed_params here (which is why I took out in the diff) was because when doing `url_for`, you have to individually pass in all the keys in `feed_params`. This is problematic because different pages have different feed params (ie, tag_pages have a `tag` param, but `/index` doesn't), so you get a complicated if statement. by doing it in jquery you eliminate this complicated logic.

knowledge_repo/app/templates/index-base.html
NiharikaRay: I responded to this above - the `feed_params` don't always have all the information required.

knowledge_repo/app/models.py
NiharikaRay: good catch! I was playing around with search fixes, forgot to take this out

NiharikaRay:
updated + responded with comments, PTAL

danfrankj:
I still suspect that there's a more elegant fix, but this should work :) LGTM

PR #152, opened 2016-11-29
enable proxy post submission via url param

danfrankj:
By going to 

`posteditor?proxy=true`

Users can add proxy posts 

image


danfrankj:
Also as a quick heads up - I don't have time to fully clean up the webeditor code so if this is a step in the right direction let's do it rather than going for fully clean code at this point.

knowledge_repo/app/templates/post_editor_markdown.html
```diff
@@ -0,0 +1,234 @@
+{% extends "post_editor_base.html" %}
```
NiharikaRay: just for my own understanding, what you did with the `html` files is to break it out from 1 template into 4 different ones yes?

danfrankj:
@earthmancash @matthewwardrop last call for comments - reaching out since this is a larger change

matthewwardrop:
This looks fine to me @danfrankj . 👍

earthmancash:
this lgtm! Will be exciting seeing the proxy posts come in

PR #158, opened 2016-12-01
excluded tags typo fixes batch tags

earthmancash:
Description of changeset: 

This one character fix gets /batch_tags up and running again:

image

Test Plan: 

ran the route locally


earthmancash:
Yup! will do

PR #160, opened 2016-12-06
Always use /feed when searching from searchbar

earthmancash:
Description of changeset: 

This fixes search when searching from viewing a post or in any other endpoint besides /feed. 

Test Plan: 

Searched from a post view


NiharikaRay:
LGTM 👍 I really need to get around to standardizing search -___-

PR #162, opened 2016-12-07
enable web publishing without approval

danfrankj:


matthewwardrop:
@danfrankj This does make it very hard to lock down instances of the knowledge repo, and I worry that it could be abused in some contexts. Perhaps it should be an option in the server configuration as to whether web posts require approval?

matthewwardrop:
I took a look at the source code for the web editor (not something I've really looked at much, that was largely the work of others). Looks like it is not locked down on the backend anyway, so this is a semantic UI change. Merging in to reflect actual backend permissions. Thanks @danfrankj

PR #166, opened 2016-12-08
check that next-btn exists before adding href in cluster view

earthmancash:
Description of changeset: 

@NiharikaRay have a look. Not sure if this is more band-aid then you'd like.

Test Plan: 

### DIFF
--- a/knowledge_repo/app/templates/index-base.html
+++ b/knowledge_repo/app/templates/index-base.html
@@ -125,7 +125,9 @@
     } else {
       next_parameters["start"] = next_parameters["results"];
     }
-    $(".next-btn")[0].href += "?" + $.param(next_parameters);
+    if ($(".next-btn").length) {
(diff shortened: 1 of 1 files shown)


knowledge_repo/app/templates/index-base.html
```diff
@@ -125,7 +125,9 @@
     } else {
       next_parameters["start"] = next_parameters["results"];
     }
-    $(".next-btn")[0].href += "?" + $.param(next_parameters);
+    if ($(".next-btn").length) {
+      $(".next-btn")[0].href += "?" + $.param(next_parameters);
```
NiharikaRay: You need to make this explicitly `== 0`, take a look at the screenshot attached
screen shot 2016-12-08 at 2 30 07 pm

NiharikaRay: actually, a better thing to do rather than checking length, would be to check for existence.

Issue #155, opened 2016-11-30
Images don't display in knowledge posts.


sryza:
Auto-reviewers: @NiharikaRay @matthewwardrop @earthmancash @danfrankj

Hi all,

I've deployed the knowledge server on top of a knowledge repo that's a MySQL DB, and I'm having trouble getting images to display in any knowledge post.

I'm running on top of master, and I've verified that the images have entries in the knowledge repo database.  For lack of a better way that I'm aware of, I've been "publishing" posts by manually updating the status of the DB entries to 3.  So it's possible that's the source of the problem?

Any idea what could be going on?  Many thanks for any guidance.

matthewwardrop:
Thanks for reporting this @sryza !

We'll look into it and see if we can reproduce it. If we can, we'll fix it; otherwise we'll try to work with you to see what's going on.

Thanks again!

ajbosco:
@sryza What are the steps for deploying a knowledge repo that's a MySQL DB?

matthewwardrop:
@ajbosco You can just do:

`knowledge_repo --repo mysql://<username>@<host>/<database>:<table> deploy`

sryza:
@matthewwardrop if I wanted to look into this myself, do you have recommendations on where in the code would be a good starting point?

earthmancash:
@sryza what does your post look like? And is it all being written in the webeditor, or are you starting from RMarkdown / iPython?

sryza:
@earthmancash I haven't used the webeditor.  The post is an IPython post with a few matplotlib-generated figures.

earthmancash:
Ok, so you're putting these iPython posts on a github repo, which the server box has locally when running the webapp? Can you tell me what the image link looks like in the knowledge.md on the git repo? Wondering if there's an issue with the path in there and where the image path is on the locally cloned git repo. (unless you're using an s3 like cloud service, in which case it's likely a different issue)

sryza:
@earthmancash I'm using the database knowledge repo, so the knowledge.md file and the image are stored as separate rows in the database.  The ref column for the image row has the value "images/output_22_1.png".  The image is referenced inside the knowledge.md as  `![png](images/output_22_1.png)`.

Does that answer your question?

earthmancash:
I think I understand. @matthewwardrop has more familiarity with what an iPython --> dbrepo setup should look like. Can you help Sandy out with where to look to debug the issue?

matthewwardrop:
[...]

PR #171, opened 2016-12-12
In DbKnowledgeRepository, better handling of parent arg in _kp_dir

sryza:
Description of changeset:
This addresses issue #155, which was causing issues not to show up in knowledge posts when using a database repository.  The change fixes the `_kp_dir` method of `DbKnowledgeRepository` to look for the parent arg in the `ref` column of the database.

Test Plan:
Without this change, images didn't display in my knowledge posts.  With the change, they did.

### DIFF
--- a/knowledge_repo/repositories/dbrepository.py
+++ b/knowledge_repo/repositories/dbrepository.py
@@ -1,6 +1,7 @@
 from builtins import object
 
 import logging
+import os
 import posixpath
 
 from sqlalchemy import create_engine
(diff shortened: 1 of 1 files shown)


matthewwardrop:
Hey @sryza! Sorry for the radio silence again. Pushing toward some rapidly approaching deadlines.

I'll do a more thorough review of the patch after Friday, but in a quick run through I have a few thoughts.

The first one is just that you are using `os.path` rather than `posixpath`; which means that on Windows there may be some inconsistencies in the way files are stored in the database. Probably worth fixing before we merge this.

I also am a bit worried about some edge cases in the modified path lookup algorithm, but it may be nothing.  Just need some time to think through the possible cases.

Will give you some more thorough thoughts once I finish the race toward EOY deadlines.

Thanks again so much for your patch and for helping us to make the Knowledge Repo amazing.

sryza:
Oops, didn't think posixpath had a relpath function, but it looks like it does.  Made that change.

knowledge_repo/repositories/dbrepository.py
```diff
@@ -204,15 +204,15 @@ def _kp_read_ref(self, path, reference, revision=None):
...
+        ref_prefix = parent + '/' if parent else ''
         revision = revision or self._kp_get_revision(path, enforce_exists=True)
         refs = (self.session.query(self.PostRef.ref)
                             .filter(self.PostRef.path == path)
+                            .filter(self.PostRef.ref.like(ref_prefix + '%'))
                             .filter(self.PostRef.revision == revision)).all()
         for (ref,) in refs:
             if ref is not None:
-                yield ref
+                yield posix.relpath(ref, parent)
```
matthewwardrop: This should be `posixpath.relpath`, not `posix.relpath`.

matthewwardrop:
Going to release a new version to fix some issues that we've been having with the knowledge_repo script, and so taking the liberty of merging this in now with my trivial fix applied.

Thanks again @sryza. We really appreciate your work! Looking forward to more patches!

sryza:
Thanks @matthewwardrop !

PR #179, opened 2016-12-19
Issue 103 dynamic template

### LINKED ISSUE #103: Template dates are hard-coded to 2016-06-29
When creating a post with knowledge_post create, it would be nice if the created_at and updated_at dates were set to the current date.

jbao:
This automatically generates the current timestamp in the template and resolves #103 

### DIFF
--- a/scripts/knowledge_repo
+++ b/scripts/knowledge_repo
@@ -12,6 +12,8 @@ import webbrowser
 import shutil
 import argparse
 import subprocess
+import tempfile
+import datetime
 from tabulate import tabulate
 
 # Register handler for SIGTERM, so we can run cleanup code if we are terminated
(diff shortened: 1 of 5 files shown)


matthewwardrop:
Hi @jbao,

Thanks for your contribution!

Can you perhaps use jinja2 templating like we are doing for the site pages itself? Thus, rather than "&current_date" you would use "{{ current_date }}", and do the substitution in the knowledge_repo script using something like:
 ```
with open('<template_name>') as f:
    result = jinja2.Template(f.read()).render(current_date=...)
```
This simplifies the code a lot.

In the testing, you can then use the `knowledge_repo create` command to do the substitution, which will also act as a good test that this mechanism is working appropriately.

Thanks again! Little fixes for nits like this are a great help, as we are often so busy with the bigger stuff that we lose sight of these sorts of things which can have a big effect on user experience.

M

**Comment by matthewwardrop** [comment]
Just thinking too: maybe you can add support for automatically populating the author too?

Should be as easy as passing in `author = os.getlogin()` to the `Template.render` method; and adding `{{ author }}` to the templates.

**Comment by jbao** [comment]
Hi @matthewwardrop , that's a good idea, will implement, btw, there're currently no test cases for the `create` command yet right?

**Comment by matthewwardrop** [comment]
@jbao How are you getting on here? Does something like the following help?
```
import pwd
import os


def get_login():
    """
    Try to retrieve the user login to in a
    """
    try:
        user_login = os.getlogin()
    except OSError:
        user_login = pwd.getpwuid(os.geteuid()).pw_name
    return user_login
```

File: scripts/knowledge_repo [hunk 1]
```diff
@@ -9,11 +9,24 @@ import threading
...
 import argparse
 import subprocess
+import datetime
 
+import jinja2
 import git
 import gitdb
 import webbrowser
 from tabulate import tabulate
+import pwd
```
naoyak, line 19: This import fails in Windows.

mdshw5:
@naoyak We're planning a deployment in our data science group, and the author and date templating are features we discussed as "nice to have", so I'm surprised to find that this functionality already exists in this pull request. Is there anything I can do to get this through the door and merged to master? Thanks!

PR #186, opened 2017-01-07
Improve rendering of knowledge posts.

matthewwardrop:
This patch improves the rendering of knowledge posts. It cleans the javascript marginally, but a proper review of the javascript code is left to future work. It also adds `pygments` to the list of packages to be installed by default, so that code highlighting works again.

Note that the author name difference is due to running locally vs. on configured server, not my improvements here.

Before:
screen shot 2017-01-07 at 9 05 29 pm

After:
screen shot 2017-01-07 at 8 57 06 pm

Not shown here are the improvements in the styling of section headers and tables.


earthmancash:
Nice, there are a bunch of things I like in this. Having a different view of the metadata, with authors underneath a centered title and emphasizing one of the created/updated dates are nice. 

Can you have it not have the white background panel though? I like the render putting everything on the default background color more.

NiharikaRay:
My only comment is to echo what Che said! I personally don't like shadows - except for pdf files that we've uploaded (a la presentation mode). 

I like the different emphasis on created_at/updated_at! This will be super useful.

matthewwardrop:
I quite like the shadows in a skeuomorphic sense, but everyone who has seen this had had the same comment... and I totally get it. Removed.

Regarding the background colour of post, though, I disagree and feel that strongly enough to put up a fight. Still willing to be outmoded, but I think there's two good reasons why the background should be white:
1) Objective: Both Python/Jupyter Notebooks and knitted R markdown files are done on a white background. This can lead to colour choices/etc which do not lend themselves well to an off-white background.
2) Subjective: I think it *looks* cleaner, and increases the contrast of the post.

If the background for the entire site was white, instead of grey, I think I'd agree with you more... but I still like the two-tone look. Attached are two screenshots highlighting the comparison of the two options.
screen shot 2017-01-09 at 2 36 08 pm
screen shot 2017-01-09 at 2 35 47 pm

Have I managed to sway anyone's mind?

jefffeng:
@matthewwardrop, I prefer the white on grey, more so than the grey on grey for the same reasons you mentioned.  That said, my top preference would be white on white as our Web DLS is guiding us towards that (e.g. https://www.airbnb.com/ and air/data).

knowledge_repo/app/static/css/custom.css
```diff
@@ -339,53 +346,116 @@ img {
...
+
 #renderedMarkdown h1,
 #renderedMarkdown h2,
 #renderedMarkdown h3,
 #renderedMarkdown h4,
 #renderedMarkdown h5,
 #renderedMarkdown h6 {
-  margin-top: 50px;
-  margin-bottom: 25px;
+  margin-top: 1.5em;
```
NiharikaRay: [nit] Is there a reason we switched from px to em?

matthewwardrop: em means the margins are relative to font sizes, and if the user changes the font sizes, this will lead to more consistent padding. It also means that we can change the font-sizes in the future, and the spacing of headers should not need to be further adjusted to look sensible.

PR #187, opened 2017-01-09
Have force_reindex also update the git repository

clabroy:
Hey there,

I'm not 100% positive this should be merged with master as it pertains to a design choice rather than a missing feature, but here's a quick fix for those looking to easily refresh the knowledge repository after a post has been added (vs. having to restart the app or set up a cron job). 

It makes the force_reindex function, accessible through the route '/debug/force_reindex', not only reindex the db, but also update the local git repository. 

The caveat is that the app must be running with --debug enabled and the knowledge_branch must be set to 'master'.

**Test Plan:** 
1. Run a local instance of the knowledge_repo with runserver and --debug enabled pointing onto a local clone of a git repository 
2. Submit a new post to the git repository from another machine
3. Visit the localhost:7000/debug/force_reindex URL and see if the git repo was properly updated and the new post is now visible

**Also, I wanted to use this opportunity to let you know that we just forked your project internally over here at Criteo!** We've been facing the exact same issues you described in your [blog post](https://medium.com/airbnb-engineering/scaling-knowledge-at-airbnb-875d73eff091#.2u1me45uy), so this tool and its use case absolutely resonated with us. 

We're starting small, but if all goes well it should become a tool widely used by our analysts. And we'll be glad to share any useful new feature we may end up coming up with. 

**Thanks a lot guys!**


matthewwardrop:
Hi @clabroy!

Exciting news to hear you have started adopting/adapting the knowledge repo to your needs. We'll gladly review any feature you think worth mainlining!

Regarding this patch in particulary: In many instances, or at least ours, the deployed server is unable to run a git pull on the knowledge data repository (due to security isolation/etc). For that reason we did not add an automatic update before reindexing, and think it probably a bad idea to accept this patch as is. We do plan, however, to add support for this in a more generic opt-in fashion, to simplify deployments (it would be rolled into the automatic index updates that are already performed on page visits if the index is older than 5 minutes). No guarantee on timelines at this point :).

Nevertheless it may be worth adding in support for this manual repository update in the mean time. I'll take another look and see if we can advise on how the update can work with different branches.

Thanks again for getting involved!

clabroy:
Hi @matthewwardrop !

I agree with your assessment, this added functionality felt a little like a quick hijack but I figured I'd post it in case it could help others.

PS-- We're now working on simplifying the post submission process for people using Windows, as it can feel like an intimidating task for those who never used bash or an Unix-like environment before. We figured the easiest way was to build a docker image for that purpose and pair it with scripts.

Thanks for the great tool :)

PR #190, opened 2017-01-11
add missing _external argument to url_for to get the full url

danfrankj:
Add missing _external argument needed in url_for to get the full URL


matthewwardrop:
Looks great. Thanks Dan!

PR #192, opened 2017-01-12
[KR] Add easy link to webeditor

NiharikaRay:
Description of changeset: If you're the author of the webpost, you now get a button on top of the rendered post to quickly take you to the webeditor
screen shot 2017-01-11 at 4 18 48 pm

Test Plan: 
1. Create a webpost where you're the author, publish it, and ensure you see the button (and the button can be clicked to take you to the webpost)
2. Create a webpost where you're not the author, and ensure the button doesn't show up. 

Auto-reviewers:  @matthewwardrop @earthmancash @danfrankj

### DIFF
--- a/knowledge_repo/app/routes/render.py
+++ b/knowledge_repo/app/routes/render.py
@@ -100,6 +100,11 @@ def render():
 
     is_author = user_id in [author.id for author in post.authors]
 
+    web_editor_prefixes = current_app.config['WEB_EDITOR_PREFIXES']
+    is_webpost = False
+    if web_editor_prefixes:
+        is_webpost = any(prefix for prefix in web_editor_prefixes if path.startswith(prefix))
+
     rendered = render_template(tmpl,
                                html=html,
                                post_id=post.id,
@@ -115,7 +120,8 @@ def render():
                                total_likes=post.vote_count,
                                tags_list=tags_list,
                                user_subscriptions=user_subscriptions,
-                               webeditor_buttons=False,
+                               show_webeditor_button=is_webpost and is_author,
+                               webeditor_buttons=is_webpost,
                                web_uri=post.kp.web_uri,
                                table_id=None,
                                is_private=(post.private == 1),
--- a/knowledge_repo/app/templates/markdown-base.html
+++ b/knowledge_repo/app/templates/markdown-base.html
@@ -15,6 +15,11 @@
             <button type="button" class="btn btn-default btn-raw">
               View Raw Markdown
             </button>
+            {% if is_author and show_webeditor_button %}
+            <button type="button" class="btn btn-default btn-webeditor">
+              View post in webeditor
+            </button>
+            {% endif %}
           </div>
         </div>
         <div class="col-md-2">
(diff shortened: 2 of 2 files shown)


knowledge_repo/app/routes/render.py
```diff
@@ -100,6 +100,11 @@ def render():
 
     is_author = user_id in [author.id for author in post.authors]
 
+    web_editor_prefixes = current_app.config['WEB_EDITOR_PREFIXES']
+    is_webeditor = []
+    if web_editor_prefixes:
+        is_webeditor = [prefix for prefix in web_editor_prefixes if prefix in path]
```
matthewwardrop: `prefix in path` is a little too loose here, since sometimes a prefix could appear within the path at some other location than the beginning; e.g.: `projects/webposts/why_you_should_use_git.kp`.

Given that you are using it as a boolean indicator, perhaps you could simplify the code here by just doing:

```
show_webeditor_buttons = any(prefix.startswith(path) for prefix in current_app.config['WEB_EDITOR_PREFIXES'])
```

danfrankj: @NiharikaRay I think we should also make sure that the current user `is_author`

knowledge_repo/app/templates/markdown-base.html
```diff
@@ -15,6 +15,11 @@
             <button type="button" class="btn btn-default btn-raw">
               View Raw Markdown
             </button>
+            {% if is_author and webeditor_buttons %}
```
danfrankj: nit: I think the server should completely control the logic of whether or not to show the button - so I'd have this bit of logic live in render.py

knowledge_repo/app/templates/markdown-base.html
```diff
@@ -89,6 +94,10 @@
   $(".btn-raw").on("click", function(){
     document.location.href = "/raw?markdown=" + encodeURI(post_path);
   })
+
+  $(".btn-webeditor").on("click", function(){
+    document.location.href = "/posteditor?path=" + encodeURI(post_path);
```
danfrankj: should we be using `url_for` ?

matthewwardrop: That's a good point. But since there are other links with the same problem, let's resolve this in a different PR.

knowledge_repo/app/routes/render.py
NiharikaRay: Way cleaner! Fixed.

knowledge_repo/app/templates/markdown-base.html
NiharikaRay: Simplified to move the boolean logic into the python file. This whole file could use some of that simplification.

knowledge_repo/app/templates/markdown-base.html
NiharikaRay: Yup, totally agree! I believe the reason that it's done this way was that it's bad practice to mix jinja templating code with JS (which I did above for some reason, and should also be fixed). We should be able to take this out and replace it with a url_for in another PR.

PR #199, opened 2017-01-17
Add updated_at to DBKnowledgeRepository database tables

danfrankj:

As discussed - adding an update field 

WARNING: if you already have a DB Repo you'll have to add a new column


knowledge_repo/repositories/dbrepository.py
```diff
@@ -3,6 +3,7 @@
 import logging
 import posixpath
 
+
```
matthewwardrop: Why did you add whitespace here?

danfrankj: accident - fixed

PR #201, opened 2017-01-18
Update urls for the online editor.

matthewwardrop:
This post updates the urls used by the web editor, which is now moved from `/post_editor?path=<path>` to `/edit/<path>`. Ajax queries are now moved under to the `/ajax/editor` and `/ajax/index` prefixes as appropriate. This patch does not attempt to fix any issues with the editor itself, but simply updates the url schema.

### DIFF
--- a/knowledge_repo/app/routes/editor.py
+++ b/knowledge_repo/app/routes/editor.py
@@ -25,30 +25,14 @@ logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
 
 
-blueprint = Blueprint('web_editor', __name__,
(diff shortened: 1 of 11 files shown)


knowledge_repo/app/routes/editor.py
```diff
@@ -241,34 +233,35 @@ def delete_post():
     return 'OK'
 
 
-@blueprint.route('/review', methods=['POST'])
+@blueprint.route('/ajax/editor/review', methods=['POST', 'DELETE'])
```
danfrankj: at some point we need to standardize review or comment but that can wait

PR #204, opened 2017-01-23
support unicode in tldr and add markdown newline extension

danfrankj:
Allow the TLDR to be unicode and fix formatting the header causing unicode errors. Also add markdown extension allowing github style md. 

### DIFF
--- a/knowledge_repo/postprocessors/format_checks.py
+++ b/knowledge_repo/postprocessors/format_checks.py
@@ -1,4 +1,5 @@
 import datetime
+from past.builtins import basestring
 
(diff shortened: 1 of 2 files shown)


knowledge_repo/postprocessors/format_checks.py
```diff
@@ -6,7 +6,7 @@
     'title': str,
     'authors': list,
     'created_at': datetime.datetime,
-    'tldr': str,
+    'tldr': (str, unicode),
```
matthewwardrop: This will break python 3. Perhaps use instead:

```
from past.builtins import basestring

...

'tldr': basestring

```

PR #205, opened 2017-01-24
dbrepo updates created at when status changes

danfrankj:


danfrankj:
no migration required and since we're not using this field elsewhere, for now, I think  this is OK

matthewwardrop:
This looks good to me; but is it worth just adding the `updated_at` column now and being done with it?

PR #206, opened 2017-01-25
on publish or unpublish, reindex even if we have checked recently

danfrankj:
@NiharikaRay @earthmancash @matthewwardrop

### DIFF
--- a/knowledge_repo/app/index.py
+++ b/knowledge_repo/app/index.py
@@ -46,15 +46,18 @@ def get_indexed_revisions():
     return indexed
 
 
-def update_index_required():
+def update_index_required(check_timeouts=True):
     if not current_app.config.get('REPOSITORY_INDEXING_ENABLED', True):
+        return False
+
+    if is_indexing():
         return False
 
     interval = current_app.config.get("INDEXING_INTERVAL", 5 * 60)  # Default to 6 minutes between indexing tasks
     seconds = time_since_index()
     seconds_check = time_since_index_check()
 
-    if is_indexing() or (seconds is not None and seconds_check is not None) and (seconds < interval or seconds_check < interval):
+    if check_timeouts and (seconds is not None and seconds_check is not None) and (seconds < interval or seconds_check < interval):
         return False
     try:
         for uri, revision in current_repo.revisions.items():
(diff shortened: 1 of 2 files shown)


knowledge_repo/app/index.py
```diff
@@ -67,23 +67,23 @@ def update_index_required():
...
 
 
-def update_index():
+def update_index(force=False):
     """
     Initialize the db from a KnowledgeRepository object
     """
 
-    if not update_index_required():
+    if not force and not update_index_required():
```
matthewwardrop: @danfrankj Perhaps we should be a little careful if an index is already underway. At the moment we have no way of cancelling an index, and so I think we should avoid starting another index if one is already happening.

`if not force and not update_index_required() or is_indexing():`

danfrankj: Makes sense  - added that logic

knowledge_repo/app/index.py
```diff
@@ -67,23 +67,23 @@ def update_index_required():
...
 
 
-def update_index():
+def update_index(force=False):
     """
     Initialize the db from a KnowledgeRepository object
     """
 
-    if not update_index_required():
+    if (not force and not update_index_required()) or (force and is_indexing()):
```
danfrankj: @matthewwardrop PTAL

matthewwardrop: @danfrankj This logic is now correct. I do think though in retrospect we should probably have just added a `skip_timeouts` option to `update_index_required`; rather than using the force option. Then we wouldn't need to check `is_indexing` multiple times. But if you want to get this in as is, we can clean up the code removing this force option in the future.

danfrankj:
@matthewwardrop that makes sense - let's do it right then :) PTAL

knowledge_repo/app/index.py
```diff
@@ -67,19 +70,19 @@ def update_index_required():
...
-    if not update_index_required():
+    if not update_index_required(check_timeouts=check_timeouts):
         return
 
     app = current_app._get_current_object()
     index_db = current_app.config['SQLALCHEMY_DATABASE_URI']
 
     if not index_db.startswith('sqlite://'):
-        t = threading.Thread(target=_update_index, args=(app,))
+        t = threading.Thread(target=_update_index, args=(app))
```
matthewwardrop: Don't remove this comma... it's needed to make this a tuple.

danfrankj: my bad - lazy mistake. Fixed

PR #207, opened 2017-01-27
[Knowledge][Bugfix] Allow tag subscription from rendered post

NiharikaRay:
Description of changeset: 
A 6 character change. With the complexity of the post javascript, the url that the subscribe button was going to was `post/<post_path>/toggle_tag_subscription`, which was throwing a 405, method not allowed error (since we only GET posts, not POST). This PR ensures that its `/toggle_tag_subscription` that gets the ajax request. 

Test Plan: 
Go to a post and subscribe to a tag from it. 

Auto-reviewers: @matthewwardrop @earthmancash @danfrankj

### DIFF
--- a/knowledge_repo/app/static/js/tags.js
+++ b/knowledge_repo/app/static/js/tags.js
@@ -96,7 +96,7 @@ function addTagSubscriptionListener(v) {
     }
     $.ajax({
       type: "POST",
-      url: url_request,
+      url: "/" + url_request,
       async: true,
       success: function() {
         // We want to toggle all the buttons for all the relevant tags


danfrankj:
Can we migrate this to use `url_for`?

NiharikaRay:
It's generally bad practice to mix Jinja and JS code. 

That being said, right now we generate all these buttons on the fly, instead of in the template. It would be a bigger change to move this to being generated in the template, but once we did that, we could use `url_for`

NiharikaRay:
The reason this is a bigger change is because we currently do the HTML formatting for the metadata in python, and tags are considered part of the metadata, so we basically "hold" a spot for them using a `<div>` and then fill that in later. Ideally, we'd do this all in the template.

danfrankj:
kk LGTM

Issue #214, opened 2017-02-06
Date not showing up in comments


NiharikaRay:
![screen shot 2017-02-06 at 1 29 10 pm](https://cloud.githubusercontent.com/assets/4268311/22667184/92c30f64-ec70-11e6-92aa-319b381fc4cd.png)

NiharikaRay:
This has been fixed: https://github.com/airbnb/knowledge-repo/pull/215

PR #215, opened 2017-02-06
[Knowledge] Add date formatting to comments

NiharikaRay:
Description of changeset: This is related to issue #214, where the date wasn't showing up in the comments. This PR changes that. 

Test Plan: 
Write a comment, note that the date now shows up in the rendered comment box.
screen shot 2017-02-06 at 1 54 39 pm

### DIFF
--- a/knowledge_repo/app/templates/markdown-rendered.html
+++ b/knowledge_repo/app/templates/markdown-rendered.html
@@ -50,7 +50,7 @@
                   <div class="row">
                     <div class="col-md-11">
                       <div class="post_comment">
-                        Posted by <b>{{ item.author }}</b> on {{ item.created_at_fmt }}
+                        Posted by <b>{{ item.author }}</b> on {{ item.format_created_at }}
                       </div>
                     </div>
                     <div class="col-md-1">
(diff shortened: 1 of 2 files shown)


knowledge_repo/app/templates/markdown-rendered.html
```diff
@@ -50,7 +50,7 @@
                   <div class="row">
                     <div class="col-md-11">
                       <div class="post_comment">
-                        Posted by <b>{{ item.author }}</b> on {{ item.created_at_fmt }}
+                        Posted by <b>{{ item.author }}</b> on {{ item.format_created_at }}
```
matthewwardrop: This definitely works, but I wonder if we should instead add a jinja filter and use that instead? Something like http://stackoverflow.com/questions/4830535/python-how-do-i-format-a-date-in-jinja2 .

NiharikaRay: Yeah - I think longterm this is probably better, but I'll merge for now and then do a bigger sweep of switching over all properties to our own jinja filters.

PR #218, opened 2017-02-08
Improve feed rendering.

matthewwardrop:
This patch set improves the rendering of the knowledge post index feed by:
 - Making rows more compact and with more consistent heights (descriptions are collapsed)
 - Visual feedback for interaction (backgrounds on hover, depression when clicked)
 - Making the who row clickable to take you to the post proper

The new javascript methods are encapsulated in a closure so that they do not clutter the global namespace, and both the javascript and css files are loaded only on the page they are required.

Below are some screenshots of the new theming:

screen shot 2017-02-08 at 11 44 19 am
screen shot 2017-02-08 at 12 24 33 pm

### DIFF
--- a/knowledge_repo/app/static/css/custom.css
+++ b/knowledge_repo/app/static/css/custom.css
@@ -98,7 +98,7 @@ body {
   color: #565a5c;
   background-color: whitesmoke;
   font-size: 14px;
-  font-family: 'Lato';
+  font-family: 'Lato', sans-serif;
   /*background-color: #F8F8F8;*/
 }
 
@@ -118,28 +118,13 @@ ul {
     margin-bottom: 10px;
 }
 
-h1 {
-  font-size: 44px;
-}
-h2 {
-  margin-top: 0;
-  margin-bottom: 0;
-  color: #484848;
-  font-size: 32px;
-  font-weight: bolder;
-}
-h3 {
-  font-size: 24px;
-}
-h4 {
-  font-size: 18px;
-}
-h5 {
-  font-size: 16px;
-}
-h6 {
-  font-size: 14px;
-}
+h1              { font-size: 2em; margin: .67em 0 }
+h2              { font-size: 1.5em; margin: .75em 0 }
+h3              { font-size: 1.17em; margin: .83em 0 }
+h5              { font-size: .83em; margin: 1.5em 0 }
+h6              { font-size: .75em; margin: 1.67em 0 }
+h1, h2, h3, h4,
+h5, h6          { font-weight: bolder }
 
 img {
     max-width:100%;
@@ -256,29 +241,6 @@ a.label:focus, a.label:hover {
   color: #9CA299;
 }
 
-.feed-author {
-  padding-bottom: 5px;
-}
-
-.feed-date {
-  color: #565a5c;
-  padding-bottom: 10px;
-}
-
-.feed-date, .feed-tldr {
-  color: #565a5c;
-}
-
-.feed-post {
-  border-bottom: 1px solid #F8F8F8;
-}
-
-.feed-post-counts {
-  color: #aaa;
-  margin-right: 3px;
-  margin-left: 15px;
-}
-
 .content-offset {
   margin-left: 50px;
 }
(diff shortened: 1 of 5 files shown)


knowledge_repo/app/static/css/custom.css
```diff
@@ -256,29 +241,6 @@ a.label:focus, a.label:hover {
   color: #9CA299;
 }
 
-.feed-author {
-  padding-bottom: 5px;
```
NiharikaRay: [comment that lingers for the entirety of the PR] The reason I named all of these fields `feed-feature` is because we call the feed, cards, and cluster all part of the "index" views. Renaming just the feed classes to index makes it more confusing in my opinion, since we have two other index views that are not going through this restyle.

knowledge_repo/app/templates/base.html
```diff
@@ -26,6 +26,7 @@
         <script src="https://cdnjs.cloudflare.com/ajax/libs/html5shiv/3.6.2/html5shiv.js"></script>
         <script src="https://cdnjs.cloudflare.com/ajax/libs/es5-shim/2.1.0/es5-shim.min.js"></script>
         <![endif]-->
+
```
NiharikaRay: [nit] unrelated change?

knowledge_repo/app/templates/index-feed.html
```diff
@@ -1,81 +1,72 @@
 {% extends "index-base.html" %}
 
-{% block inner_content %}
+{% block style_links %}
+{{ super() }}
+<link rel="stylesheet" href="{{ url_for('static', filename='css/pages/index.css')}}">
+{% endblock %}
+
+{% macro render_tag(tag, post_id) %}
```
NiharikaRay: [question] what was the reasoning for this macro? we only use it in this place, and it's made this diff a little harder to read (since the functionality hasn't changed, you've just made this a function).

knowledge_repo/app/templates/index-feed.html
```diff
@@ -1,81 +1,72 @@
...
-          </p>
-        </div>
-        <div class="col-md-6">
-          <center><img src="{{ post['thumbnail'] if post['thumbnail'] else '/static/images/default_thumbnail.png' }}" style="max-width:100%;max-height:225px" /></center>
-        </div>
-      </div>
+{% block inner_content %}
+
+{% for post in posts %}
+<div class="row index-post" data-path='{{post.path|urlencode}}' data-url="{{url_for('posts.render', path=post.path)}}">
```
NiharikaRay: same comment about `index-post` vs `feed-post`

NiharikaRay:
The hover over looks good, but the grey background on the hover panel is a little jarring. Could we change that?

knowledge_repo/app/static/css/custom.css
matthewwardrop: That's a fair point, and happy to comply. I was looking forward to the upcoming url revamp on the index routes whereby /feed becomes /index/feed (as documented earlier in an internal roadmap document). We also discussed briefly the fact that these alternative index views are potentially going to be deprecated by a more complete filtering interface, which would consolidate all of the index views into one /index route. The qualification to `index-post` instead of just `index` was supposed to allow other css classes to be present if required without clashing in the namespace. Perhaps I should have written all this in the PR description; but if you still convinced `feed-post` is a better name, I'll change them over. :).

As a side note, this won't make the diff look much different, since these css classes are defined in a separate css file (and only applied to the index/feed page to which they are relevant).

knowledge_repo/app/templates/base.html
matthewwardrop: Will remove. Made some changes here but ultimately reverted them.

knowledge_repo/app/templates/index-feed.html
matthewwardrop: The macro just made it easier for me to read the html that renders the post. Apart from that, there was a mild interest in separating form from function. Most of the structure of the post rows are now semantic, whereas the html for the tags was a little more involves a little more structure. If you prefer it all inline, happy to rearrange it.

matthewwardrop:
@NiharikaRay Will specify the background of the TLDR as an translucent color rather than a specified hex code, which should resolve this :).

knowledge_repo/app/static/css/custom.css
NiharikaRay: Got it. Given that this is pending based on future work happening, I think it probably makes sense to hold of on changing `feed-post` to `index-post`.

Fair point, i didn't realize you had re-arranged the css from one file to another.

knowledge_repo/app/templates/index-feed.html
NiharikaRay: It's fine, I was just confused as to why since the rendering didn't change.

NiharikaRay:
LGTM pending change from `index-post` to `feed-post`.

knowledge_repo/app/static/css/custom.css
matthewwardrop: Sounds good. Incoming :).

PR #220, opened 2017-02-09
Fix skipping some images based on too simple a regex.

matthewwardrop:
Description of changeset: 
Users have been reported (and I have experienced) dropping of various figures from knowledge posts. This is at least part of the problem. I've written a more sophisticated regex expression which now allows more flexible image tags (and does not group them together inadvertently).

Test Plan: 
Run on more complicated markdown scripts to ensure that the images are all collected.

### DIFF
--- a/knowledge_repo/postprocessors/extract_images.py
+++ b/knowledge_repo/postprocessors/extract_images.py
@@ -50,14 +50,18 @@ class ExtractImages(KnowledgePostProcessor):
     def find_images(cls, md):
         images = []
         images.extend(cls.collect_images_for_pattern(
-            md, '<img.*src=[\'"](.*?)[\'"].*>'))
-        images.extend(cls.collect_images_for_pattern(md, '\!\[.*\]\((.*)\)'))
+            md,
+            # Match all <img /> tags with attributes of form <name>(="<value>") with optional quotes (can be single or double)
+            # The src attribute is exected to always be surrounded by quotes.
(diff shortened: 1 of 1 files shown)


knowledge_repo/postprocessors/extract_images.py
```diff
@@ -50,14 +50,17 @@ def update_thumbnail_uri(cls, kp, images, image_mapping):
     def find_images(cls, md):
         images = []
         images.extend(cls.collect_images_for_pattern(
-            md, '<img.*src=[\'"](.*?)[\'"].*>'))
-        images.extend(cls.collect_images_for_pattern(md, '\!\[.*\]\((.*)\)'))
+            md,
+            r'<img\s+(?:\w+(?:=([\'\"])?(?(1)(?:(?!\1).)*?\1|[^>]*?))?\s+?)*src=([\'\"])(?P<src>(?:(?!\2).)*?)\2(?:\s+\w+(?:=([\'\"])?(?(1)(?:(?!\4).)*?\4|[^>]*?))?)*\s*\/?>'
+        ))
```
NiharikaRay: ooooh could we have a comment on what this regexp is supposed to match for future reference? (I'm very bad at reading regexp)

knowledge_repo/postprocessors/extract_images.py
```diff
@@ -50,14 +50,17 @@ def update_thumbnail_uri(cls, kp, images, image_mapping):
...
+        images.extend(cls.collect_images_for_pattern(md, r'\!\[.*\]\((?P<src>[^\)]*)\)'))
+        print(images)
```
NiharikaRay: do we still need to print the images?

matthewwardrop: Nope. Good catch. Removed.

knowledge_repo/postprocessors/extract_images.py
matthewwardrop: Added inline comment.

PR #221, opened 2017-02-11
Improve the pagination controls.

matthewwardrop:
Description of changeset: 
This changeset improves the styling of the feed page by adjusting margins along with a new set of pagination controls, along with another pagination bar at the bottom of the page. The pagination controls disappear if there is only one page. The code also removes the need for custom javascript on the page, and fixes a small oversight in the css for the feed hover css (the thumbnail border was not transitioning properly).

screen shot 2017-02-10 at 10 04 33 pm
screen shot 2017-02-10 at 10 04 45 pm

### DIFF
--- a/knowledge_repo/app/app.py
+++ b/knowledge_repo/app/app.py
@@ -3,6 +3,7 @@ import os
 import imp
 import logging
 import traceback
+import math
 
 from flask import Flask, current_app, render_template, g, request
 from flask_mail import Mail
@@ -10,6 +11,7 @@ from flask_migrate import Migrate
 from alembic import command
 from alembic.migration import MigrationContext
 from datetime import datetime
+from werkzeug import url_encode
 
 import knowledge_repo
 from .proxies import db_session, current_repo
@@ -153,6 +155,37 @@ class KnowledgeFlask(Flask):
                         version_revision=version_revision,
                         last_index=time_since_index(human_readable=True),
                         last_index_check=time_since_index_check(human_readable=True))
+
+        @self.template_global()
+        def modify_query(**new_values):
+            args = request.args.copy()
+
+            for key, value in new_values.items():
+                args[key] = value
+
+            return '{}?{}'.format(request.path, url_encode(args))
+
+        @self.template_global()
+        def pagination_pages(current_page, page_count, max_pages=5, extremes=True):
+            page_min = int(max(1, current_page - math.floor(1.0 * max_pages // 2)))
+            page_max = int(min(page_count, current_page + math.floor(1.0 * max_pages / 2)))
+
+            to_acquire = max_pages - (page_max - page_min + 1)
+
+            while to_acquire > 0 and page_min > 1:
+                page_min -= 1
+                to_acquire -= 1
+            while to_acquire > 0 and page_max < page_count:
+                page_max += 1
+                to_acquire -= 1
+
+            pages = list(range(page_min, page_max + 1))
+            if extremes:
+                if 1 not in pages:
+                    pages[0] = 1
+                if page_count not in pages:
+                    pages[-1] = page_count
+            return pages
 
         @self.template_filter('format_date')
         def format_date(date):
--- a/knowledge_repo/app/templates/index-base.html
+++ b/knowledge_repo/app/templates/index-base.html
@@ -11,13 +11,46 @@
   {% endfor %}
 {% endmacro %}
 
+{% macro pagination(max_pages=10, extremes=True, autohide=True) %}
+  {% if feed_params %}
+    {% set start = feed_params['start'] | default(0) %}
+    {% set results = feed_params['results'] | default(10) %}
+    {% set page = 1 if (start == 0) else (start//results + 1) %}
+    {% set page_count = feed_params['page_count'] | default(1) %}
+
+    {% if autohide and page_count > 1 %}
+      {% set page_nums = pagination_pages(current_page=page, page_count=page_count, max_pages=max_pages, extremes=extremes) %}
+
+      <div class='pagination-bar' role="group">
+        <a href="{{ "javascript:;" if (page == 1) else modify_query(start=(page-2)*results) }}" class="pagination-stepper {{ "disabled" if page == 1 }}">
+          <i class="glyphicon glyphicon-chevron-left"></i>
+        </a>
+
+        <ul class="pagination">
+          {% for page_num in page_nums %}
+            {% if loop.index0 > 0 and page_num - page_nums[loop.index0 - 1] > 1 %}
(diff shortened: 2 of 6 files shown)


danfrankj:
I didn't do an in depth review here but generally looks sane and better than the pagination we have now :) LGTM

knowledge_repo/app/app.py
```diff
@@ -3,13 +3,15 @@
 import imp
 import logging
 import traceback
+import math
```
NiharikaRay: [nit] We should alphabetize imports

knowledge_repo/app/templates/index-base.html
```diff
@@ -11,13 +11,46 @@
...
+    {% set start = feed_params['start'] | default(0) %}
+    {% set results = feed_params['results'] | default(10) %}
+    {% set page = 1 if (start == 0) else (start//results + 1) %}
+    {% set page_count = feed_params['page_count'] | default(1) %}
+
+    {% if autohide and page_count > 1 %}
+      {% set page_nums = pagination_pages(current_page=page, page_count=page_count, max_pages=max_pages, extremes=extremes) %}
+
+      <div class='pagination-bar' role="group">
+        <a href="{{ "javascript:;" if (page == 1) else modify_query(start=(page-2)*results) }}" class="pagination-stepper {{ "disabled" if page == 1 }}">
```
NiharikaRay: Why is there jinja-ified javascript as an href? We could have just modified the href in a `document.ready` call.

knowledge_repo/app/templates/index-feed.html
```diff
@@ -36,31 +36,33 @@
...
-            <span class='feed-tldr'>{{ post['tldr']|safe }}</span>
-            <span class='feed-tldr-expander'>+ Show More</span>
-            <span class='feed-stats'>
-                {% set stats = post_stats[post.path] %}
-                <span title="Users who Viewed"><i class="glyphicon glyphicon-eye-open"> </i> {{ stats['distinct_views'] }}</span>
-                <span title="Number of Favorites"><i class="glyphicon glyphicon-heart-empty"></i> {{ stats['total_likes']}}</span>
-                <span title="Comments"><i class="glyphicon glyphicon-comment"></i> {{ stats['total_comments']}}</span>
-            </span>
-            <span class="feed-dates">Created on {{ post.created_at.strftime("%B %d, %Y") }} (last updated {{ post.updated_at.strftime("%B %d, %Y") }})</span>
+<div id="feed-posts">
```
NiharikaRay: why did you wrap this in an outer div?

knowledge_repo/app/templates/index-feed.html
```diff
@@ -36,31 +36,33 @@
...
+    {% for post in posts %}
+    <div class="row feed-post" data-path='{{post.path|urlencode}}' data-url="{{url_for('posts.render', path=post.path)}}">
```
NiharikaRay: is the only diff here that everything is indented over one due to an outer div?

knowledge_repo/app/templates/index-base.html
NiharikaRay: It's generally bad practice to have javascript as an href, mixed with jinja.

knowledge_repo/app/templates/index-feed.html
matthewwardrop: Yes.

knowledge_repo/app/templates/index-feed.html
matthewwardrop: So that we can apply margins to this section. There is no standard container for the top navigation/buttons, so this was the (perhaps temporary) solution.

NiharikaRay: couldn't you have added this here: https://github.com/airbnb/knowledge-repo/pull/221/files/b9c8ff670213a6a7350609b706e35dfe48e167d2#diff-1ed009af666446201bba775cdde98ef4R24

You could have added a `top-pagination-bar` and `bottom-pagination-bar` class (or some way to differentiate the two), to add the padding.

knowledge_repo/app/templates/index-base.html
matthewwardrop: In general I agree about using javascript in hrefs, but in this case I felt it is worth it, and I'll note there is no javascript code modified by jinja here.

The href is a javascript void function rather than `#` because that way it does nothing rather than sending people to the top of the page, which was the previous behaviour (and annoying).

If the page is anything other than 1, then the link is pure html and simply the current query modified such that start query parameter is offset as appropriate. Yes, this parsing *can* be done in javascript, but I felt it simpler to do in Python.

Do you agree?

knowledge_repo/app/app.py
matthewwardrop: Yeah... we should go through and do this consistently where it makes sense (here for example).

knowledge_repo/app/templates/index-feed.html
matthewwardrop: Indeed I could. I was mirroring our behaviour on the rendered markdown page where #renderedMarkdown is used. Adding containers for the controls would require separate blocks to be set up... and then I'd want to roll it out to all derived pages, which would have conflated multiple changes into one patchset... confusing things.

Perhaps I should roll such a suggestion into the upcoming PR adding sidebars?

knowledge_repo/app/templates/index-base.html
NiharikaRay: I do not agree.

> The href is a javascript void function rather than # because that way it does nothing rather than sending people to the top of the page, which was the previous behaviour (and annoying).

Correct. That being said, you could easily do something like `<a href=# onClick='return false'>`, which would prevent the default behavior. 

In my opinion, if you wanted to do this in Python, then you can precompute the links server-side, and send a dictionary through, with the jinja template doing something like 
`<a href={{ page_url[page]}}` and then have the optional  `onClick` attribute depending on whether it was the first page or not. 

Obviously, however, pre-computing it is non-ideal, which is why I believe you should actually do it in Javascript, on pageload (similar to how it was done). You would then avoid the `onClick` attribute issue.

knowledge_repo/app/templates/index-feed.html
NiharikaRay: 👍

knowledge_repo/app/templates/index-base.html
matthewwardrop: Oh nice... I didn't release that `onclick='return false'` prevents the default event bubbling. That is a better way to do it.

It would have been cleaner to do something like:
```
<a href="{{ "#" if (page == 1) else modify_query(start=(page-2)*results) }}" class="pagination-stepper {{ "disabled" if page == 1 }}" {{ 'onclick="return false"' if page == 1 }}>
```

PR #225, opened 2017-02-15
Fix Windows deployment related code

matthewwardrop:
This fixes two related issues:
 - The `create` action in the knowledge_repo script did not terminate upon succeeding, allowing subsequent code to run, which included importing of the deployment code. On the the Windows platform, this resulted in an error importing gunicorn (gunicorn does not support non UNIX platforms).
 - A hard import of GunicornDeployer would, on Windows platforms, make it impossible to "deploy" the server. I do not imagine Windows servers being used often in this capacity, but it is worth preventing this spurious behaviour.


danfrankj:
LGTM

PR #234, opened 2017-02-27
Add new table for storing error logs, as distinct from pageviews.

matthewwardrop:
This patch-set adds a separate error log table instead of combining error logs with the pageviews table. This will be used in my next PR to store index updating errors also; and can be used by other internal functions in the future.

In the migration, past error logs are dropped. I don't foresee this being an issue... but if necessary, we can fix the migration to conserve historical error logs. My take is that these logs are not worth maintaining, and will lead to incomplete error logs being stored in the error log table (due to us now collecting more information when errors occur).


danfrankj:
Nice! LGTM

PR #235, opened 2017-02-27
add proxy posts as an option

danfrankj:
Simple change to make proxy posts an option.

@matthewwardrop 
@NiharikaRay 

image


matthewwardrop:
LGTM. Am planning to revamp this page shortly, since I will be adding support for uploading a standalone/portable knowledge post... but this can definitely get in there before I get around to it :).

PR #241, opened 2017-03-11
allow author to see posts even if they are private

danfrankj:
Also do we still need to have implicit groups 

@NiharikaRay @matthewwardrop 

```
        # create an implicit group, group_post.id, to add
        # single users to
        group = Group(name=":post_group_" + str(self.id))
```


danfrankj:
@NiharikaRay can you comment?

danfrankj:
@NiharikaRay ping!

PR #253, opened 2017-04-10
Fix plotly, tags and 'Hide All Code' button for markdown

tanni3009:
Description of changeset: 
Change jquery selectors for renderedMarkdown (was changed from id to class), so that tags and 'Hide All Code' button work again
Add plotly.js as static module and place it before required.js to fix plotly charts in markdown

Test Plan: 

### DIFF
--- a/knowledge_repo/app/static/js/pages/index-feed.js
+++ b/knowledge_repo/app/static/js/pages/index-feed.js
@@ -5,7 +5,13 @@
     }
 
     function posts_open(event) {
-        window.location = $(this).data('url');
+        var is_mac = navigator.platform.indexOf('Mac') != -1;
+        if (is_mac && event.metaKey || !is_mac && event.ctrlKey) {
+            window.open($(this).data('url'), '_blank');
+        } else {
+            window.location = $(this).data('url');
+        }
+
     }
 
     function posts_expand_tldr(event) {
--- a/knowledge_repo/app/templates/base.html
+++ b/knowledge_repo/app/templates/base.html
@@ -17,7 +17,7 @@
         <script src="{{ url_for('static', filename='modules/select2/js/select2.min.js')}}"></script>
         <script src="{{ url_for('static', filename='modules/hightlight.pack.js/highlight.pack.js')}}"></script>
         <script src="{{ url_for('static', filename='modules/marked.js/marked.js')}}"></script>
-        <!-- <script src="{{ url_for('static', filename='modules/plotly-latest.js/plotly-latest.js')}}"></script> -->
+        <script src="{{ url_for('static', filename='modules/plotly-latest.js/plotly-latest.js')}}"></script>
 
         <!-- require js is used for plotly, but has a bunch of collisions with other js packages
              make sure to have it be last js package imported -->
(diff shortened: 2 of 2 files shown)


matthewwardrop:
Hi @tanni3009,

Thanks for taking the time to submit a patch to the knowledge repo! The name id->class changes definitely make sense, but including plotly.js (which is a fairly hefty module) indiscriminantly for every post may not be the best way to approach this (especially since some versions are incompatible, and we want old posts to be able to render too). I think by default the plotly offline package includes all the javascript required into the output itself, and hence only needs `require` to be in the javascript namespace. Is that right?

tanni3009:
Hi @matthewwardrop,

Thanks for the answer! You are right, including plotly.js is probably not the best approach. 
I tried to get it to work with require, but to be honest, I've never worked with it before so I'm stuck now. Could you give me some guidance on how to do it?

Breaka84:
Hi @matthewwardrop,

Thanks for your input!
Unfortunately, including the js-code right before the plotly-graph does not work ('Plotly not defined').
To make the markdown-file itself leaner, we normally just include the div without plotly.js.
Maybe it's possible to make loading plotly.js optional? I think about setting a tag while deploying a post, to include plotly.js (before require.js). This would keep backwards compatibility with old posts and allow us to load plotly.js as a static module if needed.

matthewwardrop:
Hey @Breaka84 and @tanni3009,

I took a look into this over the weekend and it looks like the javascript/plotly issues were related to the way that the markdown was converted to HTML. If fixed that issue (and in the process improved the rendering of the cell blocks and outputs, which included the toggle buttons) in #256 . At this stage, I think everything should be working as expected (including Plotly graphs). If it works in the Jupyter notebook, it should work in the knowledge repo.

If you get a chance, can you test this, and if everything is fixed to your satisfaction, feel free to close this PR :).

Best,
M

Breaka84:
Hi Matthew,

thank you for your effort!
Unfortunately, it still doesn't work. It's the same error message: `Plotly is not defined`...
But this PR can be closed as the `tags` and `hide code` button are now fixed!
We (my company) can live with adding the plotly.js statically by using my fork.

Greetings,
David

matthewwardrop:
Hi @tanni3009 @Breaka84,

Just tested the knowledge_repo code again, and plotly seems to work great. Perhaps I'm missing something? This is something we want to support, but given that plotly.offline.init_notebook_mode(connected=True/False) will link to plotly js on a cdn / embed javascript in notebook, I don't think you need to patch knowledge_repo to achieve this. I added a sample knowledge post with Plotly figures here: https://github.com/airbnb/knowledge-repo/blob/master/tests/test_posts/plotly.ipynb .

Let me know if I can be of any further assistance!

M

Breaka84:
Hi @matthewwardrop,

I see now where the root of our misunderstanding lies :-)
Yes, plotly in a jupyter notebook (ipyn) without a hitch, but we are using and having problems with Plotly graphs within simple markdown files (md). I'll try to come up with a simple example to demonstrate the problem.

Best,
David

Breaka84:
Hi @matthewwardrop,

I uploaded some test files which show the problem we were describing with plotly within simple markdown-files at
https://github.com/Breaka84/knowledge-repo/tree/plotly_md_test/tests/test_posts/plotly
Namely `simple_plotly_markdown_with_included_plotly_js_code.md`, `simple_plotly_markdown_with_online_polotly_js.md` & `simple_plotly_markdown_without_plotly_js.md`

Best,
David

Issue #238, opened 2017-03-10
support for javascript within a post


bmabey:
Auto-reviewers: @NiharikaRay @matthewwardrop @earthmancash @danfrankj

Nearly all of our IPython notebooks use [Bokeh](http://bokeh.pydata.org/en/latest/) which will load some custom javascript. When the notebook is rendered to markdown the JS loading gets lost in translation and the inserted HTML by bokeh is rendered as a code block:

![image](https://cloud.githubusercontent.com/assets/1837/23802505/d49baa5e-0570-11e7-9d28-f53d3579026d.png)

Is this a known issue/limitation? If so is there a current workaround or a fix in progress?

matthewwardrop:
Thanks for reporting this! We weren't fully aware of this. We'll look into this soon :).

matthewwardrop:
Should be fixed in #256. Tested with Plotly, and everything looks good. If this still doesn't work for you @bmabey, please reopen the issue :).

bmabey:
Thanks @matthewwardrop! 

Are there any docs on how to get up and running on `master` so I can try this out? I've tried using pip install  git+https://github.com/airbnb/knowledge-repo.git` and `python setup.py install` but both result in `knowlege_repo` script in an un-runnable state. I get following message when I try to run it:

```
Traceback (most recent call last):
  File "/Users/bmabey/anaconda/bin/knowledge_repo", line 4, in <module>
    __import__('pkg_resources').run_script('knowledge-repo==0.7.6', 'knowledge_repo')
  File "/Users/bmabey/anaconda/lib/python3.6/site-packages/pkg_resources/__init__.py", line 738, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/Users/bmabey/anaconda/lib/python3.6/site-packages/pkg_resources/__init__.py", line 1491, in run_script
    raise ResolutionError("No script named %r" % script_name)
pkg_resources.ResolutionError: No script named 'knowledge_repo'
```

Any thoughts?

matthewwardrop:
Hi @bmabey,

This is an odd error, but if worst comes to worst you can do: `./scripts/knowledge_repo` within the checkout out knowledge-repo git repository. I wonder if this has something to do with anaconda? You could also try installing it into a virtual environment too :).

PR #259, opened 2017-05-09
adds support for bokeh

bmabey:
Description of changeset: 
Following up on #238 and #256 this changesets updates Knowledge Repo to work with ipython notebooks that have bokeh. Two changes were required to make this happen.
 * Support for cell types that had mime type of `application/javascript`. The default markdown converter for `nbconvert` ignores these types.
 * Some special case handling of script/div tags that get emitted by bokeh in HTML cells (see commit for more info).

Test Plan: 
All tests were passing for me but no new automated tests were added. Rather, an example notebook with `bokeh` was added similar to the `plotly` one.


knowledge_repo/converters/ipynb.py
```diff
@@ -1,11 +1,23 @@
...
 from .._version import __optional_dependencies__
 
 TEMPLATE = '''
 {%- extends 'markdown.tpl' -%}
 
+{%- block data_javascript scoped %}
+{% set div_id = uuid4() %}
+<div id="{{ div_id }}"></div>
+<div class="output_subarea output_javascript {{ extra_class }}">
+<script type="text/javascript">
```
matthewwardrop: Is this `element` variable expected to be present by notebook javascript cells? Or is this simply for debug purposes?

knowledge_repo/converters/ipynb.py
```diff
@@ -1,11 +1,23 @@
 import os
+import nbconvert
```
matthewwardrop: Importing `nbconvert` here adds a hard dependency on `nbconvert`, which we would prefer to avoid. We also don't need it here, since we never using anything directly from the `nbconvert` package namespace.

knowledge_repo/converters/ipynb.py
```diff
@@ -46,13 +58,17 @@ def from_file(self, filename):
...
+                                                 'image/svg+xml', 'text/latex', 'image/png', 'image/jpeg',
+                                                 'text/plain']
 
         nb = nbformat.read(filename, as_version=4)
 
         dl = DictLoader({'full.tpl': TEMPLATE})
         md_exporter = MarkdownExporter(config=c, extra_loaders=[
                                        dl], template_file='full.tpl')
         (body, resources) = md_exporter.from_notebook_node(nb)
+        body = body.replace('</div>\n<script', '</div>\n\n<script')
```
matthewwardrop: Why is this necessary? Once it is rendered in HTML, why do we need multiple newlines?

knowledge_repo/converters/ipynb.py
bmabey: This was an accident on my part, sorry. I've removed this import and fixed-up the relevant commit.

knowledge_repo/converters/ipynb.py
bmabey: Cells that have `bokeh` plots are HTML cells that look like:

```html
    <div class="bk-root">
        <div class="bk-plotdiv" id="0b1810b9-8707-4646-87a4-598bfb2fbc21"></div>
    </div>
<script type="text/javascript">
...
```

The downstream processor in the knowledge repo processes the script tag
improperly in this case. Meaning that it doesn't treat the `script` tag as HTML and so the plot does not render. I found that by putting a newline between the div and the
script the script tag is treated as HTML and the plot appears. You can reproduce this problem quickly by commenting out this line and adding the bokeh notebook to a repo. You'll see that bokeh loads but that no plots are rendered in the HTML view (they do appear in the Markdown view).

A more elegant solution is probably to fix the underlying issue in the Markdown processer downstream so that `script` tags appearing afer `div` tags are still treated as HTML. I tried to do just that by modifiying the `IndentsAsCellOutputProcessor` which you introduced in #256. I didn't fully understand that code and I was unable to fix the issue. So, I punted and fixed it upstream.

knowledge_repo/converters/ipynb.py
bmabey: What `element` are you referring to?  Do you mean the `extra_class` ? This block of code was taken from the `nbconvert` template for HTML. We are free to change this of course but I felt like we should render it as closely as `nbconvert` does for HTML. It could be that `extra_class` gets pulled from cell metadata (total guess!) before the template is rendered.

matthewwardrop: The `element` variable I'm talking about is: on line 14 (the one under this annotation as it turns out... whoops!
`var element = $('#{{ div_id }}');`

knowledge_repo/converters/ipynb.py
matthewwardrop: Thanks for the explanation. I tracked this down to a limitation of the markdown html collecting preprocessor, and have fixed it in #263. If you rebase on master, you should be able to drop this workaround.

bmabey: Very nice! I verified that this fixes the problem so I have rebased onto master and removed this commit.

knowledge_repo/converters/ipynb.py
bmabey: My guess is that it is for debugging if anything (again, I copy and pasted it from https://github.com/jupyter/nbconvert/blob/68a3925efef5c304c2c8682ad05cac9560c4173a/nbconvert/templates/html/basic.tpl#L202-L211).  I removed it and and the bokeh plots still work so we can probably remove it if that is your preference.

matthewwardrop: Okay. That makes sense. Since it is included upstream, we'll keep it here too. Thanks!

PR #263, opened 2017-05-10
Fix cell output globbing preventing html tags from being collected by markdown preprocessors.

matthewwardrop:
This fixes some of the workarounds needed by #259 .

### DIFF
--- a/knowledge_repo/converters/html.py
+++ b/knowledge_repo/converters/html.py
@@ -34,6 +34,28 @@ MARKDOWN_EXTENSTIONS = ['markdown.extensions.extra',
                         'knowledge_repo.converters.html:IndentsAsCellOutput']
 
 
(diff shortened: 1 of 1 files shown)


knowledge_repo/converters/html.py
```diff
@@ -34,6 +34,31 @@
...
+            if not line.startswith(' ' * self.markdown.tab_length):
+                if in_block:
+                    if line != "":
+                        lines.insert(i, "")
+                    in_block = False
+                else:
+                    if line == "":
+                        block_startable = True
+                    else:
+                        block_startable = False
```
NiharikaRay: [nit] could probably simplify the following logic block into a one liner

matthewwardrop: That's true. Would make it easier to read too :). Will fix before merging. Thanks Nikki.

PR #264, opened 2017-05-11
Further improve post theming.

matthewwardrop:
This patch adds some theming improvements that I've had sitting around for a while, rounding out some of the previous improvements. In particular, it:

- Fixes rendering of non-HTML cell output (now again rendered in monospace fonts)
- Removes automatic shrinking of code blocks to 400px
- Hides code collapsing buttons until code block is hovered, and then it shows to the right of the cell with a glyph.
- Renders code blocks and their output in a flatter style.
- Adds subtle drop shadow to page
- Moves some markdown rendering specific CSS to a dedicated file that is only added to the relevant pages.

Notes: The resize icon is done using UTF-8 characters, since glyphicons are not working. When the HTML refresh is done, this will be updated to use a glyphicon.

screen shot 2017-05-10 at 9 59 03 pm

On mouseover of top cell block:
screen shot 2017-05-10 at 9 59 19 pm


danfrankj:
Hey @matthewwardrop! I'm not designer but the new look seems fine to me. Also good fixes for hiding code.

PR #268, opened 2017-05-16
Add orgmode support

arinarmo:
Description of changeset: 

Adds support for [orgmode](http://orgmode.org/) documents. 

We decided to keep the orgmode [export settings](http://orgmode.org/manual/Export-settings.html) syntax instead of requiring a YAML header, so that `#+TITLE: the title` gets translated to `- title: the title`. 

In the cases where there is no convention for an orgmode document (e.g. the `updated_at` field of the metadata YAML), we defined as prefixed by KR_, for knowledge repo (so `KR_UPDATED_AT`). 

I would especially like your thoughts on that.

Test Plan: 


matthewwardrop:
Thanks for your contribution @arinarmo! Hitherto I have never heard of orgmode files, but this sounds like a cool converter to have. Moreover, apart from a few idiomatic things, your submission looks pretty tidy!

A few questions:
 - Does it work? Can you add a sample orgmode document to the `test_posts` folder under `tests` so that tests of this converter can be automated?
 - Would it make sense to use a Python library like https://github.com/bjonnh/PyOrgMode rather than doing all the string substitutions yourself? (I'm not sure how mature this library is... it was just a random search result)
 - Can we name the converter `orgmode.py` instead of `org.py`?
 - Prefixing knowledge repo specific headers with 'KR_' makes sense to me.... but does this follow any precedents in the orgmode community?

Once we sort these issues out, I'll do a more fine-grained code review, and then we can get this merged!

arinarmo:
I'm actually pretty new to orgmode myself, some people in our team like it (especially since you can use it as a multi-language notebook with the [babel plugin](http://orgmode.org/worg/org-contrib/babel/))  and since I really want to use your project, I thought it would be nice to write support for it.

Answering your questions:
* I'll add a test in the next couple of days
* This makes a lot of sense and would make the code much cleaner, will look into it
* I'm following your convention by naming the converter after the file extension, as opposed to the full name (`rmd.py` instead of `rmarkdown.py`). I understand simply naming it `org.py` can be confusing though, Do you feel an exception to the convention is called for?

For the last question, a brief search yielded nothing. If you feel this is important, I can ask around in the mailing list. 

However, I found orgmode has defined export settings for `DESCRIPTION` and `KEYWORDS` which I think loosely map to `tldr` and `tags`, then the only metada field left would be `KR_UPDATED_AT`, which I think is optional. So, every required YAML field can be mapped to an already defined orgmode setting, and the optional ones set to `KR_OPTIONAL_FIELD`. Thoughts?

arinarmo:
Perhaps a better solution (read: with precedent) for the metadata question is to put everything in a single property, [like babel](http://orgmode.org/worg/org-contrib/babel/how-to-use-Org-Babel-for-R.html), this would look like:

`#+KNOWLEDGE_REPO: :updated_at <date> :thumbnail <number>`

while keeping every field that has a equivalent standard org field (so `TITLE` and `AUTHOR` are still there)

matthewwardrop:
Ah yes... `org.py` is more consistent, and so I retract my suggestion :).

Regarding metadata formats, your most recent suggestion sounds good to me, since this would also allow the orgmode document to be more easily (and clearly) used by different interpreters :).

matthewwardrop:
Hi @arinarmo ,

Are you still interested in merging this work?

arinarmo:
Hello @matthewwardrop ,

Definitely still interested, only I haven't got time to replace the substitutions with a more mature python library that does the .org parsing. I'll do my best to look into it this week.

csharplus (approved):
LGTM.

PR #281, opened 2017-06-13
Docker for testing

zerogjoe:
This PR:
- Dockerizes the knowledge_repo application, using an image based on the current Travis build environment
- Updates the Travis build/test script to use the dockerized application
- Adds a system test container to check that the application is being served correctly (this consists of a stub for now with just one test to check that GET /feed returns a page)
- Deploys a Docker image to Docker Hub, providing that the environment variables DOCKER_USERNAME, DOCKER_PASSWORD and DOCKER_REPO are configured in Travis
- Adds a requirements.txt so that prerequisites can be installed in a Docker layer without running setup.py (improves caching of Docker layers during development)
- Adds environment variables PYPI_USER, PYPI_PASS_SECURE and PYPI_REPO to make the deploy process to PyPI configurable in Travis (and prevent deploy being triggered from an outside fork like this one!)

Auto-reviewers: @NiharikaRay @matthewwardrop @earthmancash @danfrankj

PR #299, opened 2017-07-06
Cloud git integration

zerogjoe:
Hi all - this branch is a first pass at an authentication framework for Knowledge Repo, with a working authenticator implemented for Bitbucket OAuth.

(Some details of our implementation of the Knowledge Repo - our remote is being hosted on a private Bitbucket repo, and we are using Bitbucket both for storage and SSO authentication. The Knowledge Repo app itself is running on an Azure App Service for Linux container, via Docker.)

This branch also contains the features from my previous Docker for Testing PR: https://github.com/airbnb/knowledge-repo/pull/281

Key changes are as follows:

* Added the Flask-Login UserMixin class as a parent class to models.User.
* Added the extensible KnowledgeRepositoryAuthenticator class from which to build service-specific authenticators
* Added the NoCheckAuthenticator class as a subclass of KnowledgeRepositoryAuthenticator - the default authenticator type, which merely checks for the user's username from HTTP headers and logs them in
* Added the BitbucketAuthenticator class as a subclass of KnowledgeRepositoryAuthenticator - uses Bitbucket's OAuth API to authenticate the user, and check that they have permissions to see the git remote before serving knowledge posts
* Modifications to app.py to integrate the Flask-Login machinery
* Updated unit tests for the above

The standard authentication flow is that the user will hit these four endpoints in order:

/auth/before_login - does any necessary prep work, and redirects to /auth/login
/auth/login - does the actual redirect to the authentication provider
/auth/authorized - receives the callback from the authentication provider
/auth/after_authorized - performs any necessary checks with the newly authorized API, and authenticates the user with Flask-Login

The Bitbucket implementation also contains this additional endpoint:

/auth/gitwebhook - target for post-push webhook from Bitbucket to tell the app to update its local copy of the repo

Accessing any other endpoint other than /auth/* will check whether the user is authenticated with Flask-Login, and if not, redirect to /auth/before_login.


matthewwardrop:
Thanks for the contribution @zerogjoe !

You beat me to this by a couple of days, and it looks like the approach you've taken is very similar to the one that we were going to take. I'll try to take a look over this PR tonight, and give you some feedback.

One quick note... it looks like you've modified the newline character everywhere in the files you've edited. I'm guessing that you are doing your work on Windows?

zerogjoe:
You're welcome! Let me know if you have any questions. And yes, I do work on Windows, and that's the nudge I needed to fix the line endings issue on my dev environment (it's been driving me nuts because it breaks Bash scripts!)

matthewwardrop:
@zerogjoe I had a closer look through your patches, and there's a lot of great stuff here. Thanks so much for taking the time to make the knowledge repo work better for yourself (and everyone). While we had plans on leveraging flask-login, and had a pretty good idea what we wanted to support, you've made that very concrete.

With that said, though, we want to be pretty careful in how we implement this since we don't want to change it again in the foreseeable future. Rather than do a back and forth where you try and guess what's in our heads as we figure it out ourselves, do you mind if we take the code contributed here related to authentication, learn from it, and then create a new patch set which we're happy supporting into the indefinite future? I suspect you'll recognise a lot of whatever we contribute in the end. Needless to say, you'll be thoroughly recognised for your work.

I'll comment on the Docker-related code in your other PR.

zerogjoe:
@matthewwardrop That works best for me too! Looking forward to seeing the finished version.

matthewwardrop:
Excellent! Thanks @zerogjoe !

matthewwardrop:
This work was eventually superseded in #305 . Thanks again for your help @zerogjoe !

Issue #133, opened 2016-11-09
Add separate documentation for Windows installations?


RMHogervorst:
Auto-reviewers: @NiharikaRay @matthewwardrop @earthmancash @danfrankj

matthewwardrop:
This is probably a good idea. Do you have some expertise setting up Windows environments for this sort of thing? We fixed the code to work on Windows, but are not sure how the community at large deals with Windows installations.

RMHogervorst:
Unfortunately my only experience is with my own computer.

Op 10 nov. 2016 07:54 schreef "Matthew Wardrop" <notifications@github.com>:

gstoel:
Here also someone who is willing to help/test stuff out to get this working on windows. 
currently it fails to run on windows with ImportError: No module named fcntl (which I think is due to the fact that gunicorn is not supported under windows). 

@matthewwardrop you mention you have fixed the code to work on Windows right? Unfortunately not at ' my Windows'  :) so willing to help get it working there and feed back to the community.

RMHogervorst:
I'm willing to test but I'm a Python novice. So I'll not take the lead here

Op 15 nov. 2016 6:02 p.m. schreef "Geoffrey Stoel" <notifications@github.com

gnatamania:
If it's possible to get this to work on windows I would love to see some documentation on this! I've been trying to step through the code and fix myself to no avail :(

bdoohan:
@matthewwardrop I have come across the same issue, seeing "ImportError: No module named 'fcntl'" when I try to preview my post. Has anyone developed a fix for this?

bdoohan:
@matthewwardrop @earthmancash To follow up, I have come across the same issue, seeing "ImportError: No module named 'fcntl'" when I try to preview my post. Has anyone developed a fix for this?

matthewwardrop:
@bdoohan This is fixed in #225. Whenever the deployment code was run, it would trigger this fault due to Gunicorn only supporting POSIX platforms. There was a related bug for the "create" action, which did not terminate the knowledge_repo script after a successful template creation, triggering the above bug also.

matthewwardrop:
@gstoel @RMHogervorst @gnatamania Apart from the bug fixed above, are there any other issues that you experience?

gstoel:
@matthewwardrop I will retry the installation later this week... will get back to you with details or other issues.

RMHogervorst:
I will try the installation today and report back @matthewwardrop

[...]

PR #302, opened 2017-07-10
Automatically replace windows image paths with POSIX paths when base64-encoding images

zerogjoe:
A very simple patch for when image paths in markdown files get generated in the Windows style (backslash) as opposed to the POSIX style (forward slash).

Specifically, this is an issue when submitting .Rmd files from Windows. (Mentioned in the comments for this issue: https://github.com/airbnb/knowledge-repo/issues/133)


susanjo80:
Hi,

any idea when this will be available in the master branch?

Tx,
S.

matthewwardrop:
@susanjo80 @zerogjoe The right place to do this transformation is probably earlier in the workflow, rather than just before rendering. I'll take a look at this over the weekend.

matthewwardrop:
This was fixed in #315 for all new posts. If you need to migrate old posts, you can manually replace the image urls in the markdown, or write a script to do so. Happy to help if necessary :).

PR #305, opened 2017-07-17
Add support for new types of authentication

matthewwardrop:
**WARNING: This PR requires a migration of your database to function, which is not yet finalised. Please do not run the code in this PR against a production database.**

This PR adds support for new types of authentication, including OAuth2.0. It also adds support for new user attributes like profile photos. While work is still underway to fully flesh out these new features, they do "work". Many thanks to @zerogjoe for his work scoping out this kind of work in #299, and proving out some of these concepts!

**Progress in this PR:**
- [x] Add a flexible abstract interface (`KnowledgeAuthProvider`) for administrators to add custom authentication providers
  - [x] Support arbitrary `User` identifier and field extraction
  - [x] Support arbitrary validation of user privileges
- [x] Extend `User` class to support new properties
  - [x] Rename existing `username` field to `identifier` (identifiers are what will be used to reference authors in post headers, usernames will be the raw name used to log in, which may be different across authentication providers; ensuring that identifiers are consistent across different authentication providers is the responsibility of server administrators).
  - [x] Re-add `username` and add `password` fields for local login support (not fully implemented in this PR)
  - [x] Finalise naming of new fields like `avatar_uri`, `email`, and `name`.
- [x] Add some reference / default `KnowledgeAuthProvider` implementations
  - [x] Debug login (no username/password checking performed)
  - [x] BitBucket  (OAuth 2.0) [needs testing + potential improvement]
  - [x] GitHub (OAuth 2.0)
  - [x] Google (OAuth 2.0)
- [x] Improve server-side configuration by inheriting from a default configuration
- [x] Add user-facing UI for new authentication process
  - [x] Add automatic route addition for authentication providers
  - [x] Add profile avatar into navbar (will be elsewhere too)
  - [x] Add sign in and sign out actions to navbar when appropriate
  - [x] Add new login switcher
- [x] Implement configurable policy for which endpoints require authentication
- [x] Create database migration
- [x] Fix unit-tests that are broken (if any) by these changes.

**Not in scope (but compatible with this PR):**
 - Full local login management (registration, local storage of username/password/ etc).
 - LDAP/Kerberos login implementation

And for the kicks, here's a screenshot of the new authentication screen (if more than one auth provider is configured).
screen shot 2017-07-22 at 11 03 21 pm


zerogjoe:
I like what you've done here @matthewwardrop! The user_info_mapping concept is particularly smart and solves a problem I was having with my implementation.

One thing that I think is missing is the ability in the OAuth authentication flow to check whether the user has access to certain specific resources before granting access. You'll want to do this in most use cases, as it's very rare that you'll want to grant access to all Github or Google users. For example, in our Bitbucket implementation we do this:

```
        # if repository has a Bitbucket remote, check that we can access it or deny access
        if self.app.repository.git_has_remote:
            git_remote_url_parts = self.app.repository.git_remote.url.split(':')
            if git_remote_url_parts[0] == "git@bitbucket.org":
                resp = bitbucket.get("repositories/" + git_remote_url_parts[1].split('.')[0])
                if resp.status_code != 200:
                    return redirect(url_for("auth.denied"))
```

Do you think there's a way to generalise this behaviour across OAuth providers, or whether the best way to achieve this is by subclassing OAuth2Provider to extend it for specific use cases?

matthewwardrop:
Hi @zerogjoe ,

Sorry... I should probably have commented on this aspect of the process.

So far, I have been considering the authentication process to be on a server by server instance, rather than a backend repository by repository instance. I think this makes sense given that you want to be consistent in author attribution over all repositories attached to a server instance. Unfortunately, this does mean there is no way to handle specific per-repository authentication in a generic manner. The `KnowledgeRepository` instance attached to a Knowledge Repository server does not need to be a git repository, but could be a backed by a database, or stitched together from multiple sources.

Instead, to cater for the use case you describe, there is the `validate` method (which gets called after the user object is extracted, and before being logged in. You can then validate that the user should be given access to the website and/or (in the future, once this is implemented) attach different roles to the user. I was contemplating allowing one to pass in this validation function to the `KnowledgeAuthProvider` constructor as well, but it seemed cleaner in those cases to just have the system administrator subclass it for their needs and repository configuration. Note that in this class you have access to the resolved `User` object and also to the `self.oauth_client` object, and so can do further digging to resolve any required privileges.

What are your thoughts @zerogjoe ?

zerogjoe:
That works well. Exposing it in the server config is a nice solution!

In theory, should this branch be ready for testing? If so, I can try to set aside some time later this week to build an instance of our server that uses this patch set.

matthewwardrop:
Yep. This code works, with the caveat that you should *not* connect it to your existing databases. The code will just fail to work, due to certain fields being missing, but particularly once I start adding the migration code, I'd prefer it be kept away from production systems until such time as it is ready to be merged.

Testing the BitBucket auth would definitely be helpful. I've tested the debug, Google and GitHub authentication providers; but don't personally have a bitbucket account. It would also be good to have you prove the system in your environment, just to prove that there is no critical feature missing.

I should have proper roles and permissions implemented by next week.

matthewwardrop:
@danfrankj @NiharikaRay @zerogjoe This is now ready to merge. There's a little more to be done in exposing permission limitations in a more user-friendly way (if users do not have appropriate permissions, which basically means they are logged out atm, then they can attempt to do things that silently fail), but this patch is large enough as is.

knowledge_repo/app/app.py
```diff
@@ -4,52 +4,67 @@
...
+
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
 
 
 class KnowledgeFlask(Flask):
 
-    def __init__(self, repo, db_uri=None, debug=None, config=None, **kwargs):
+    def __init__(self, repo=None, db_uri=None, debug=None, config=None, **kwargs):
```
NiharikaRay: [question] why does this have a default value?

knowledge_repo/app/models.py
```diff
@@ -234,26 +235,59 @@ class Vote(db.Model):
...
+
+    can_logout = True
+
+    # Other useful methods
     @property
     def format_name(self):
-        username_to_name = current_repo.config.username_to_name
-        return username_to_name(self.username)
+        return self.preferred_name or self.name or self.identifier
+        # username_to_name = current_repo.config.username_to_name
```
NiharikaRay: [question] how will this work with existing implementations that have a `username_to_name` function for formatting usernames? Are you planning on running a migration to backfill these columns?

knowledge_repo/app/models.py
```diff
@@ -234,26 +235,59 @@ class Vote(db.Model):
...
     __tablename__ = 'users'
 
     id = db.Column(db.Integer, primary_key=True)
-    username = db.Column(db.String(500))
     created_at = db.Column(db.DateTime, default=func.now())
 
+    identifier = db.Column(db.String(500))  # Unique identifier across all login methods
+
+    username = db.Column(db.String(500))  # Username used to log in (may differ from identifier)
+    password = db.Column(db.String(500))  # Password for local logins
```
NiharikaRay: [question] Should we encrypt these passwords before storing them in the db?

knowledge_repo/app/models.py
```diff
@@ -234,26 +235,59 @@ class Vote(db.Model):
...
+
+    name = db.Column(db.String(500))  # Name as determined by auth method
```
NiharikaRay: [question] What is the difference between `identifier`, `username`, and `name`?

knowledge_repo/app/models.py
```diff
@@ -275,7 +309,7 @@ def get_subscriptions(self):
         return out_subscriptions
 
     @property
-    def get_liked_posts(self):
+    def liked_posts(self):
```
NiharikaRay: [nit] Changing the name of this function seems unrelated to this diff - for the future, can we separate renames from core functionality (this diff is really large as is).

knowledge_repo/app/models.py
```diff
@@ -234,26 +235,59 @@ class Vote(db.Model):
...
     def format_name(self):
-        username_to_name = current_repo.config.username_to_name
-        return username_to_name(self.username)
+        return self.preferred_name or self.name or self.identifier
+        # username_to_name = current_repo.config.username_to_name
+        # return username_to_name(self.username)
 
     @property
-    def get_subscriptions(self):
+    def subscriptions(self):  # TODO: make attribute style naming
```
NiharikaRay: [nit] Changing the name of this function seems unrelated to this diff

knowledge_repo/app/auth_provider.py
```diff
@@ -0,0 +1,78 @@
+import sys
+import re
+from abc import abstractmethod
+from builtins import object
+
+# from .app.models import User
```
NiharikaRay: [nit] If you don't need this, delete?

NiharikaRay: [edit2] especially since you import it below

knowledge_repo/app/auth_providers/oauth2.py
```diff
@@ -0,0 +1,165 @@
...
+    }
+}
+
+
+def _resolve_oauth_config(name, local_namespace, config, *variables):
+    presets = PRESETS.get(name, {})
+    output = []
+    for variable in variables:
+        value = local_namespace.get(variable, None)
+        if value is None:
```
NiharikaRay: [nit] Can we not do something like `if not value`?

knowledge_repo/app/auth_providers/oauth2.py
```diff
@@ -0,0 +1,165 @@
...
+        )
+
+    def prompt(self):
+        return redirect(self.oauth_client.authorization_url(self.authorization_url)[0])
+
+    def get_endpoint_url(self, endpoint):
+        return urljoin(self.base_url, endpoint)
+
+    def get_user(self):
+        code = request.args.get('code')
```
NiharikaRay: [question] Do we want a default value here? Or are we okay passing in None to `fetch_token`

knowledge_repo/app/routes/auth.py
```diff
@@ -0,0 +1,56 @@
...
+
+    if len(providers) == 1:
+        return redirect(url_for('auth_provider.{}.prompt'.format(providers[0].name)))
+
+    return render_template(
+        'auth-switcher.html',
+        providers=[{'name': provider.name, 'icon_uri': provider.icon_uri, 'link_text': provider.link_text} for provider in providers]
+    )
+
+# @blueprint.route('/auth/login/<provider>', methods=['GET', 'POST'])
```
NiharikaRay: [nit] Delete commented code?

knowledge_repo/app/routes/posts.py
```diff
@@ -68,16 +70,18 @@ def render(path):
 
     comments = post.comments
     for comment in comments:
-        comment.author = db_session.query(User).filter(User.id == comment.user_id).first().username
+        author = db_session.query(User).filter(User.id == comment.user_id).first()
+        if author is not None:
```
NiharikaRay: [nit] `if not author` I believe should be able to suffice here

tests/test_comments.py
```diff
@@ -53,6 +52,8 @@ def test01_comment_on_a_post(self):
 
             assert rv.status == '200 OK'
 
+            print(rv.data)
```
NiharikaRay: delete

tests/test_webeditorposts.py
```diff
@@ -58,6 +58,8 @@ def test01_webeditor_route(self):
 
             assert rv.status == '200 OK'
 
+            print(rv.data)
```
NiharikaRay: [nit] delete.

knowledge_repo/app/app.py
matthewwardrop: This leaked through from some other things I am/will be working on. Good catch.

knowledge_repo/app/models.py
matthewwardrop: I was thinking we would hash them using sha256 or something. I suppose we *could* encrypt them, now that we have secret key; if we want to be able to recover the underlying password in the future... but I think that hashing them is more standard practice. If we plan to encrypt, then it may be necessary to increase the size of the password field column; whereas with hashing, 500 is more than generous :).

knowledge_repo/app/models.py
matthewwardrop: **identifier** is the username/email/identifier used to indicate the authors of posts in the YAML headers. Exactly what it is (email, LDAP username, some combination) is left up to the server administrator to decide. My recommendation for a default would be the author's email, since that means that posts can be handed around between different knowledge repo instances and have very little chance of clashing with other users; however, I was hesitant to force that on people. Internally we've been using LDAP usernames as the identifier.

Do you think we should force identifiers to be emails? That could be good for the knowledge post ecosystem in the long term.

**username** is the username provided to the Auth provider, which may be different than the **identifier**, especially if deployers set up multiple authentication methods. (i.e. You can log in with Google OR with LDAP.) It is mainly here as a way to allow administrators to have a *local* login method that does not require external logins, but where the username does not match the identifier.

**name** is the literal name of the person. I would be "Matthew Wardrop", or "M Wardrop" or "Matthew"; depending on how the authentication method chooses to set this field. The **preferred_name** is a preference field for users to override this default (not currently exposed in UI, since this PR was big enough), which is especially useful in places where the authentication provider cannot properly render names (such as CJK names or unusual non-Latin names).

knowledge_repo/app/models.py
matthewwardrop: The configuration for the repository and the configuration for the server interface do not **have** to line up, since the server interface can get a lot more information about users than is possible in the repository itself.

So far, we've lifted the configuration for names in the server directly from the primary repository, but this was actually only ever a stop-gap for this more complete authentication system... which can now get *actual* names as verified by various authentication providers. I think repositories will still need an `identifier_to_name` method because they are not privy to insight from the authentication providers.

I will do the updating of the repository side of things in a different PR, since this one is already huge; and delete that commented out code.

knowledge_repo/app/models.py
matthewwardrop: It is unrelated. This too slipped through from general improvements. My apologies.

knowledge_repo/app/models.py
matthewwardrop: As above.

knowledge_repo/app/auth_provider.py
matthewwardrop: That's a pretty good reason to delete it ;). Good catch. Thanks Nikki.

knowledge_repo/app/auth_providers/oauth2.py
matthewwardrop: In this case, there is a difference between things like `[]`, `False`, `''` and `None`, which may vary between OAuth providers. If any "False-like" object is passed that is not `None`, it should be the value passed on.

knowledge_repo/app/auth_providers/oauth2.py
matthewwardrop: A `None` value for `code` will almost always result in an error, but since this is OAuth provider specific, it is possible that it won't. If it does result in an error, then an Exception of the appropriate type will be thrown, which is also nice (we don't need to pre-emptively assert anything).

knowledge_repo/app/routes/auth.py
matthewwardrop: Yep. In my haste to be done with the patch, I left some unused sample code in there. Thanks for picking it out.

knowledge_repo/app/routes/posts.py
matthewwardrop: I think you're right, here... but the fact that I cannot tell you with 100% certainty off-the-cuff troubles me. The problem is that Python offers a whole heap of magic behaviours, depending on which methods are provided, and if anyone upstream of us decides that it would be a good idea to have SQL models (or other complex datatypes) be iterable, then doing a `not <object>` operation would lead to slower and possibly incorrect code. I prefer to be defensive against upstream changes by being more explicit in such comparisons, especially for more complex Python datatypes, so I don't have to worry about implementation details, or new weird bugs appearing in the future.

If you feel strongly about it though, I think it should be safe enough here. The crew managing sqlalchemy are probably to do anything crazy.

tests/test_comments.py
matthewwardrop: Roger!

tests/test_webeditorposts.py
matthewwardrop: Whoa. I left *two* debugging statements in. Thanks Nikki!

matthewwardrop:
Agreed regarding testing plans.

The good news regarding the groups functionality is that they're next from an authentication work perspective. I haven't included it in this PR because it is huge already... but groups will get merged with this authentication work; allowing auth-providers to specify groups (e.g. LDAP), and server adminstrators to decide whether custom groups can be made beyond those provided by authentication providers. This will then be linked in with principal (the user role/permissions management framework used here), and so become a first tier feature of the authentication system. If there are bugs, we'll find them pretty soon (certainly before we release v0.8.0).

knowledge_repo/app/models.py
matthewwardrop: Oh... And I should add that the `username` and `password` fields here are **only** used if an authentication provider wants to do **local** authentication. Typically the values for both of these fields will be null.

knowledge_repo/app/app.py
```diff
@@ -30,26 +36,35 @@ def __init__(self, repo, db_uri=None, debug=None, config=None, **kwargs):
...
+        assert isinstance(self.repository, knowledge_repo.KnowledgeRepository), "Invalid repository object provided."
 
-        # Lift debug mode from underlying KnowledgeRepository object if not specified
-        self.config['DEBUG'] = debug if debug is not None else repo.config.debug
+        # Set debug mode from kwargs or else maintain current setting
+        if debug is not None:
+            self.config['DEBUG'] = debug
+
+        # Set the secret key for this instance (creating one if one does not exist already)
+        self.config['SECRET_KEY'] = self.config['SECRET_KEY'] or str(uuid.uuid4())
```
danfrankj: asking to learn: what's this for?

knowledge_repo/app/routes/auth.py
```diff
@@ -0,0 +1,36 @@
...
+    )
+
+
+@blueprint.route("/auth/logout")
+@login_required
+def logout():
+    logout_user()
+
+    # Notify flask principal that the user has logged out
+    identity_changed.send(current_app._get_current_object(),
```
danfrankj: asking to learn: what is _get_current_object

knowledge_repo/app/auth_providers/oauth2.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def extract_user_from_api(self):
+
+        def extract_from_dict(d, key):
+            if isinstance(key, (list, tuple)):
+                if len(key) == 1:
+                    key = key[0]
+                else:
+                    return extract_from_dict(d[key[0]], key[1:])
+            elif isinstance(key, str):
```
zerogjoe: Pretty sure you want `if isinstance(key, str):` here for the desired behaviour. As is, passing in any tuple or list as key results in `RuntimeError: Invalid key type: (name of last string in list or tuple).`

knowledge_repo/app/auth_providers/oauth2.py
```diff
@@ -0,0 +1,165 @@
...
+    'bitbucket': {
+        'base_url': 'https://api.bitbucket.org/2.0/',
+        'authorization_url': 'https://bitbucket.org/site/oauth2/authorize',
+        'token_url': 'https://bitbucket.org/site/oauth2/access_token',
+        'auto_refresh_url': 'https://bitbucket.org/site/oauth2/access_token',
+        'scopes': 'account',
+        'user_info_endpoint': 'user',
+        'user_info_mapping': {
+            'identifier': 'username',
+            'name': 'name',
```
zerogjoe: Should be `'name': 'display_name',`

knowledge_repo/app/auth_providers/oauth2.py
```diff
@@ -0,0 +1,165 @@
...
+            'auto_refresh_url',
+            'scopes',
+            'user_info_endpoint',
+            'client_id',
+            'client_secret',
+            'user_info_mapping',
+            'validate'
+        )
+        if validate is not None:
+            self.validate = validate
```
zerogjoe: This implementation implies functions in the server_config.py in the format:

`def OAUTH_{provider name}_VALIDATE(user):`

whereas what we really need is 

`def OAUTH_{provider name}_VALIDATE(provider, user):`

to be able to make use of further oauth_client_get calls.

Could this be replaced by something like `self.validate = lambda x: validate(self, x)`?

knowledge_repo/app/app.py
matthewwardrop: The secret key is used by flask-login to encrypt cookies so that it's not possible to modify authentication details.

knowledge_repo/app/routes/auth.py
matthewwardrop: For `werkzeug.LocalProxy` objects, all methods and attributes are lazily proxied. For some reason, the reference guides for flask-principal have you get the actual underlying object (which is retrieved using the `_get_current_object`). I'm not sure why this is required, or if it is... but it may be to guarantee that the object exists in the current runtime.

knowledge_repo/app/auth_providers/oauth2.py
matthewwardrop: Thanks for point this out and your PR.

knowledge_repo/app/auth_providers/oauth2.py
matthewwardrop: As above :).

knowledge_repo/app/auth_providers/oauth2.py
matthewwardrop: Yep. I didn't realise that binding to self was different then binding to cls. :). Thanks!

shashj199:
@matthewwardrop 

Hey, I am really excited about the authentication feature and will love to test it out. Is there any kind of documentation available on how to implement it?

Thanks

matthewwardrop:
@shashj199 You should simply be able to add a provider (such as `'github'`) to the `AUTH_PROVIDERS` server configuration. See the `config_defaults.py` file in `knowledge_repo/app`. You will also need to configure an OAuth client id/secret on the authentication server-side.

shashj199:
@matthewwardrop Thanks a lot, will definitely try it out.

PR #315, opened 2017-08-07
Fix reference path in Windows.

matthewwardrop:
In Windows, reference paths should still be indicated using posix/html-like path separators, otherwise images break for posts generated on Windows. This simple patch resolves this for all *new* posts. (This supersedes that #302 that added a hack later in the post rendering pipeline, and which could on occasion break urls containing `\`).

@zerogjoe @susanjo80: Can you verify that this resolves these issues for you?

### DIFF
--- a/knowledge_repo/postprocessors/extract_images.py
+++ b/knowledge_repo/postprocessors/extract_images.py
@@ -1,4 +1,5 @@
 import os
+import posixpath
 import re
 import logging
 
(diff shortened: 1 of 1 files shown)


susanjo80:
I receive an "AttributeError: module 'posixpath' has no attribute 'path'"
Any idea what's up with that?

matthewwardrop:
@susanjo80 The code `posixpath.path` does not exist anywhere in this codebase... so I'm a little perplexed. Can you paste the traceback as well?

susanjo80:
Here you go :)
image

matthewwardrop:
@susanjo80 Did you manually apply my patch, instead of changing the git branch? Because that's not the code on this branch. `posixpath.path.join` != `posixpath.join`.

susanjo80:
This worked! :)

PR #316, opened 2017-08-07
Update source paths from `orig_src` to `src`.

matthewwardrop:
With the current workflow, users are not supposed to directly modify the markdown, and so we expect that the source code should be in sync with the rendered format. This tight coupling will continue to strengthen in upcoming versions (with new tooling), so to avoid confusion, we here rename `orig_src` to `src` inside of posts. This patch maintains legacy accessing of `orig_src` paths for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,7 +176,7 @@ See the file itself for more detail.
 
 #### Knowledge Web Application Configuration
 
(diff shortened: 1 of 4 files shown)


danfrankj:
LGTM

Issue #243, opened 2017-03-16
Rendering posts with embedded images is very slow


rene-giretzlehner:
Auto-reviewers: @NiharikaRay @matthewwardrop @earthmancash @danfrankj

we are running knowledge-repo in our company since a few weeks on a server and are facing issues with high cpu load and very slow response times when clicking on a knowledge post the first time. It takes up to 20sec to render the post. During this time the CPU is up to 100%. When clicking on the post a 2nd time - it's fast as you would expect a page to render. The same thing occurs with every other post as well. 
As soon as we restart the server process - the issues start from scratch.

Did you ever have any of these issues? We're running on a ubuntu box.

matthewwardrop:
Hi @rene-giretzlehner ! Thanks for reporting this issue to us!

When the knowledge-repo starts up, it creates/updates an index database for the posts. If you do not specify a path for this index, it creates the index in-memory. I suspect that that is happening is that you have not specified a path, so every time it restarts it needs to recreate this index.

Can you confirm that this is the case?

To prevent this initial load, you can use a local sqlite or remote (potentially shared) database between (perhaps more than one) knowledge_repo instances. For example:

`knowledge_repo --repo <repository> deploy --dburi sqlite:///test.db`
`knowledge_repo --repo <repository> deploy --dburi mysql://db_user@localhost/knowledge`

Does this help?

rene-giretzlehner:
We have configured the database via server_config.py
`SQLALCHEMY_DATABASE_URI = 'sqlite:////home/deep/knowledge.db'`
and startup the server with 
`knowledge_repo --repo {magic}/home/deep/deepthought --repo {webposts}sqlite:////home/deep/dbrepo.db:mypostreftable runserver --config server_config.py --port 80`

So the database grows over time and viewcount etc. updates. But on every server restart the behaviour is again as described in our initial post.

matthewwardrop:
@rene-giretzlehner Hmmm... I think I understand what is going on here. Currently, the database sync is triggered if it has been more than 5 minutes since the last index and the data repository has changed since its last update. The SQLite database does not support threading, and so when using an SQLite database to store the index metadata, updates are done synchronously. There are improvements in the pipeline for these synchronisation processes, not all of which will reach sqlite databases unfortunately, you can greatly reduce this initial startup time by:
1) Using `deploy` rather than `runserver` to deploy the knowledge repo. The `runserver` mode is primarily for development, and does not handle multiple requests very well.
[...]

PR #322, opened 2017-08-24
adding local image extraction

aaronbiller:
Description of changeset: Adds a postprocessor to extract images to a local directory

This module follows the same pattern as `extract_images_to_s3.py` but copies the files to a local directory instead.  By extracting images to a centralized directory in the repository and running an http-server through `npm`, we were able to solve our troubles with slow and failing posts (see #243).  A server could also be set up with python.

Here's the method we're going to attempt:
- Run both the `extract_images_to_s3` and `extract_images_to_local` postprocessors locally
- Add the `'static-server'` directory to `.gitignore` so images aren't committed to the repository
- When deploying to production, sync the s3 bucket to the same `'static-server'` directory and start an http server that will serve the images from that directory

This process will keep images out of the repository, keep all images synced in s3, allow users to preview their posts locally when contributing, and decouple serving images from the rest of the work the knowledge repo is doing on the production server.


resources/extract_images_to_local.py
```diff
@@ -0,0 +1,71 @@
...
+                repo_name=repo_name,
+                time=int(round(time.time() * 100)),
+                random_string=random_string,
+                ext=img_ext).strip().replace(' ', '-')
+
+            # See if a static file directory exists, if not, let's create
+            if not os.path.exists(LOCAL_HTTP_SERVER_DIRECTORY):
+                os.makedirs(LOCAL_HTTP_SERVER_DIRECTORY)
+            # Copy images to local http server directory
+            cmd = "cp {} {}/{}".format(
```
matthewwardrop: This assumes a UNIX filesystem in a UNIX environment. Perhaps you can change this to use `os.path.join` and `shutil.copyfile` instead of hardcoding the '/' and the `os.system` command?

resources/extract_images_to_local.py
```diff
@@ -0,0 +1,71 @@
...
+            retval = os.system(cmd)
+            if retval != 0:
+                raise Exception('Problem copying images')
+        finally:
+            # Clean up temporary file
+            if is_ref:
+                os.remove(tmp_path)
+
+        # return uploaded path of file
+        return posixpath.join(LOCAL_HTTP_SERVER_ROOT, fname_img)
```
matthewwardrop: Perhaps you can use `urlparse.urljoin` (python 2) and `urllib.parse.urljoin` (python 3) here, instead of posixpath?

aaronbiller:
@matthewwardrop Thanks for reviewing!  I'll go ahead and make these changes from my side.  I also have a python http server module started that can be used to serve images from a specified directory.  I can include that here or start a separate PR with that code when it's ready.  Obviously there are other solutions like `npm http-server` or `python -m SimpleHTTPServer` that are perfectly functional, but not a bad idea to have everything self-contained for ease of use.

To answer your question about #243, we're using Python 2.7, but switching to 3.6 had no effect.

sbourke:
Can this be merged ?

aaronbiller:
@matthewwardrop changes were made, do we have the green light here?

aaronbiller:
Happy New Year! Coming back around to our implementation of knowledge_repo... can we merge this in?  We've been using this configuration for a few months now internally.
@matthewwardrop @NiharikaRay @earthmancash @danfrankj

matthewwardrop:
@aaronbiller @sbourke Many apologies! The past 6 months have been very busy! This is indeed ready to be merged. Merging it in now.

PR #324, opened 2017-09-02
Improve indexing by separating timing into separate thread

### LINKED ISSUE #295: Add support for fetching updates to git repositories during index update
After deploying the knowledge_repo in a server we've seen that it needs to be restarted to display the new content pushed to the repo. Is this the expected behavior? Are we missing something? Thanks

matthewwardrop:
... and making use of improved error logging.

This patch reworks how index synchronisation is done. Previously, every thread would attempt to trigger an update to the index (if needed) at request time. This patch changes this behaviour such that one instance/thread of the knowledge repo claims to be the index master, and remains so until it fails to update the index; whereupon this responsibility automatically shifts to a difference instance. This should greatly reduce the chance of lock collisions.

As a result of this improved logic, we can now update the knowledge repositories as part of the index update, for which initial support is added in this PR. This will likely be improved on the actual post storage side in the future, to enable, for example, tunnelled git operations for updating `GitKnowledgeRepository` instances. As such, this PR addresses and closes #295.

It also dumps errors in the synchronisation process into the errorlog, making it much easier to diagnose errors.

@jdavidheiser: Can you check that this works in your set up?

### DIFF
--- a/knowledge_repo/app/config_defaults.py
+++ b/knowledge_repo/app/config_defaults.py
@@ -139,6 +139,44 @@ def prepare_repo(repo):
 
 
 # ---------------------------------------------------
+# Repository Indexing configuration
+# ---------------------------------------------------
+# The Knowedge Repo updates the index of available posts on a regular basis.
+# If the database is not thread-safe (i.e. in the case of SQLite), then the
+# index will be updated on the main thread before every request that is more
+# than `INDEX_INTERVAL` seconds after the last sync completed. Otherwise,
+# indexing will occur every `INDEX_INTERVAL` seconds after the previous sync.
+# Syncing is designed to be compatible with multiple instances of the Knowledge
+# Repo connected to the same database, accross multiple machines and/or
+# processes; and so a global indexing lock is employed. When a sync begins,
+# a sync lock is put in place and the responsible process is considered to be
+# the primary agent responsible for syncing until its last update is longer than
+# `INDEXING_TIMEOUT` seconds, whereby the lock is ceded to the next requesting
+# process. Note that `INDEXING_INTERVAL` must be larger than `INDEXING_TIMEOUT`
+# or strange things might begin to happen.
+INDEXING_INTERVAL = 5 * 60  # 5 minutes
+INDEXING_TIMEOUT = 10 * 60  # 10 minutes
+
+# Whether an index operation should update repositories
+INDEXING_UPDATES_REPOSITORIES = True
+
+# Whether repositories should be updated even without a sync lock (in which case
+# the repositories will be updated on the sync timers, even if the relevant
+# process/thread does not have a lock on updating the index). This is useful in
+# context of multiple Knowledge Repo servers working together to serve the
+# repositories across multiple machines, which each require repository syncing.
+# Disable this if (for some reason) you have multiple Knowledge Repo servers
+# running on the same machine, and you want to avoid potential clashes. This
+# key is ignored if `INDEXING_UPDATES_REPOSITORIES` is False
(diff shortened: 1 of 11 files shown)


jdavidheiser:
I just confirmed that this does work on our internal deployment of Knowledge Repo.  Thanks for the quick response on this!

matthewwardrop:
Great! Thanks @jdavidheiser !

knowledge_repo/app/config_defaults.py
```diff
@@ -116,6 +116,44 @@ def prepare_repo(repo):
...
+# index will be updated on the main thread before every request that is more
+# than `INDEX_INTERVAL` seconds after the last sync completed. Otherwise,
+# indexing will occur every `INDEX_INTERVAL` seconds after the previous sync.
+# Syncing is designed to be compatible with multiple instances of the Knowledge
+# Repo connected to the same database, accross multiple machines and/or
+# processes; and so a global indexing lock is employed. When a sync begins,
+# a sync lock is put in place and the responsible process is considered to be
+# the primary agent responsible for syncing until its last update is longer than
+# `INDEXING_TIMEOUT` seconds, whereby the lock is ceded to the next requesting
+# process. Note that `INDEXING_INTERVAL` must be larger than `INDEXING_INTERVAL`
```
danfrankj: typo: INDEXING_TIMEOUT

knowledge_repo/app/index.py
```diff
@@ -19,8 +24,58 @@
...
+    if app.check_thread_support(check_index=True, check_repositories=app.config['INDEXING_UPDATES_REPOSITORIES']):
+        if os.environ['KNOWLEDGE_REPO_MASTER_UUID'] != app.uuid:
+            logger.info("Not spawning index-sync timers for non-master application instance: {}".format(app.uuid))
+            return
+
+        def index_watchdog(app):
+            while True:
+                if not hasattr(app, 'sync_thread') or not app.sync_thread.is_alive():
+                    logger.warning("Master indexing thread has died. Restarting...")
+                    context = None
```
danfrankj: what does this do?

knowledge_repo/app/index.py
```diff
@@ -19,8 +24,58 @@
...
+            IndexMetadata.set('lock', 'master_check', current_app.uuid)
+            db_session.commit()
+            if (
+                IndexMetadata.get('lock', 'index_master') == current_app.uuid or
+                time_since(IndexMetadata.get_last_update('lock', 'index_master'), default=current_app.config["INDEXING_TIMEOUT"] + 1) > current_app.config["INDEXING_TIMEOUT"]
+            ):
+                IndexMetadata.set('lock', 'index_master', current_app.uuid)  # Update/set lock and update timestamp
+                return True
+        finally:
+            IndexMetadata.set('lock', 'master_check', None)
```
danfrankj: I'm a little confused by the master_check. regardless of what happens in the `try:` statement, don't we immediately set the master_check back to none?

knowledge_repo/app/config_defaults.py
matthewwardrop: Thanks! Fixed.

knowledge_repo/app/index.py
matthewwardrop: Nothing. An entirely vestigial piece of code. Thanks!

knowledge_repo/app/index.py
matthewwardrop: That's correct... but it means that only one thread can attempt to become the master indexer at any time, preventing race conditions that may allow two processes to both be the master indexer for the first subsequent sync operation. It's unlikely to happen, but coding defensively.

PR #329, opened 2017-09-21
debug authentication providers

danfrankj:
@matthewwardrop 

 * Add missing imports so that authentication providers don't reference undefined variables 
 * Add messaging for when a user reaches a resource that requires authentication but no authentication providers are configured 
 * Remove unused imports 
 * Fix inconsistent handling of avatar images  
 * Add email and login for github authentication provider to be used as identifier
 * Add exceptions when authentication providers are not configured with client secret / key or when autherntication providers do not return the requried information

### DIFF
--- a/knowledge_repo/app/auth_providers/oauth2.py
+++ b/knowledge_repo/app/auth_providers/oauth2.py
@@ -1,12 +1,7 @@
 import json
-from flask import request, render_template, flash, redirect, url_for
-from flask_login import login_user
+from flask import request, redirect
 from six.moves.urllib.parse import urljoin
-
 from ..models import User
-from ..proxies import db_session
-from ..utils.auth import is_safe_url
-
 from ..auth_provider import KnowledgeAuthProvider
 
 
@@ -21,7 +16,7 @@ PRESETS = {
         'user_info_mapping': {
             'identifier': 'username',
             'name': 'display_name',
-            'photo_uri': ['links', 'avatar', 'href']
+            'avatar_uri': ['links', 'avatar', 'href']
         }
     },
     'github': {
@@ -32,9 +27,9 @@ PRESETS = {
         'scopes': None,
         'user_info_endpoint': 'user',
         'user_info_mapping': {
-            'identifier': 'email',
+            'identifier': ['email', 'login'],
             'name': 'name',
-            'photo_uri': 'avatar_url'
+            'avatar_uri': 'avatar_url'
         }
     },
     'google': {
@@ -50,7 +45,7 @@ PRESETS = {
         'user_info_mapping': {
             'identifier': 'email',
             'name': 'name',
-            'photo_uri': 'picture'
+            'avatar_uri': 'picture'
         }
     }
 }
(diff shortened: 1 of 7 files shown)


knowledge_repo/app/templates/base.html
```diff
@@ -104,18 +106,6 @@
             </li>
           </ul>
 
-          <!-- Expose the same links as above in the collapsed navbar header -->
```
danfrankj: what is this for?!

knowledge_repo/app/auth_providers/oauth2.py
```diff
@@ -32,9 +27,9 @@
         'scopes': None,
         'user_info_endpoint': 'user',
         'user_info_mapping': {
-            'identifier': 'email',
+            'identifier': 'login',
```
matthewwardrop: I have a preference that the identifier always be emails where possible, since it makes sharing of knowledge posts between instances a sane experience, and allows for collaborations between different institutions.

knowledge_repo/app/templates/base.html
```diff
@@ -84,15 +84,17 @@
           <ul class="nav navbar-nav navbar-right visible-sm-block visible-md-block visible-lg-block">
             <li class="dropdown">
               <a href="#" class="avatar-dropdown dropdown-toggle" data-toggle="dropdown" role="button" aria-haspopup="true" aria-expanded="true" style='padding: 5px;'>
-                  <img class="avatar" src="{% if current_user.is_anonymous %}{{url_for('static', filename='images/profile_photo_logged_out.png')}}{%else%}{{ current_user.avatar_uri or url_for('static', filename='images/profile_photo_default.png') }}{%endif%}" />
+                  <img class="avatar" src="{% if current_user.is_anonymous %}{{ url_for('static', filename='images/profile_photo_logged_out.png') }}{%else%}{{ current_user.avatar_uri or url_for('static', filename='images/profile_photo_default.png') }}{%endif%}" />
                   <span class="caret"></span>
               </a>
               <ul class="dropdown-menu">
-                {% if current_user.is_authenticated %}
+                {% if not current_user.is_anonymous and current_user.is_authenticated %}
```
matthewwardrop: I don't think you can be authenticated *AND* be anonymous.

knowledge_repo/app/templates/base.html
matthewwardrop: This is for the responsive menus. You don't want to delete it :).

matthewwardrop:
This looks mostly good to me (see my comments). Apart from renaming the `photo_uri` to `avatar_uri`, which makes it more consistent (thanks), and removing some unnecessary imports (something I was deferring to a separate PR for clarity), and doing some sanity checks on server side configuration regarding OAuth tokens, does this actually fix a bug? What needed to be debugged?

knowledge_repo/app/templates/base.html
danfrankj: kk - I'll have to ask you about that later but putting this back

knowledge_repo/app/auth_providers/oauth2.py
danfrankj: Updated to email, login as users don't have to have an email for github. Specifically, my email comes back as None (even though github certainly has it?)

danfrankj:
The main bugs here were that some missing imports caused undefined variables thus making it impossible to use the authentication providers. I'd recommend we release a new version after we merge this.

danfrankj:
@matthewwardrop updated in response to your comments and added a more explanatory summary of this changeset to the description of this PR PTAL

Issue #332, opened 2017-10-10
Deploy on Heroku


alexisrosuel:
Auto-reviewers: @NiharikaRay @matthewwardrop @earthmancash @danfrankj

I am trying to host my knowledge repo on Heroku, and I have the same issue than [issue 314](https://github.com/airbnb/knowledge-repo/issues/314) . This issue has been closed because the author found another solution (using Docker), but I would like to reopen this issue in order to fix it directly.

As a reminder, here is the Procfile to deploy on Heroku :
`web: knowledge_repo --repo /app runserver --port $PORT`

and here is the error log :
```
2017-08-04T18:53:26.093397+00:00 app[web.1]:   File "/app/.heroku/python/bin/knowledge_repo", line 228, in <module>
2017-08-04T18:53:26.093587+00:00 app[web.1]:     repo = knowledge_repo.KnowledgeRepository.for_uri(args.repo)
2017-08-04T18:53:26.093590+00:00 app[web.1]:   File "/app/.heroku/python/lib/python3.6/site-packages/knowledge_repo/repository.py", line 46, in for_uri
2017-08-04T18:53:26.093687+00:00 app[web.1]:     return cls._get_subclass_for(scheme)(uri, *args, **kwargs)
2017-08-04T18:53:26.093688+00:00 app[web.1]:   File "/app/.heroku/python/lib/python3.6/site-packages/knowledge_repo/repository.py", line 76, in __init__
2017-08-04T18:53:26.093788+00:00 app[web.1]:     self.init(**kwargs)
2017-08-04T18:53:26.093790+00:00 app[web.1]:   File "/app/.heroku/python/lib/python3.6/site-packages/knowledge_repo/repositories/gitrepository.py", line 66, in init
2017-08-04T18:53:26.093912+00:00 app[web.1]:     self.path = self.uri.replace('git://', '')
2017-08-04T18:53:26.093914+00:00 app[web.1]:   File "/app/.heroku/python/lib/python3.6/site-packages/knowledge_repo/repositories/gitrepository.py", line 88, in path
2017-08-04T18:53:26.093998+00:00 app[web.1]:     path), "Provided path '{}' is not a valid repository.".format(path)
2017-08-04T18:53:26.094002+00:00 app[web.1]: AssertionError: Provided path '/app' is not a valid repository.
```

Thanks

danfrankj:
Hey @alexisrosuel! Thank you for your interest in the knowledge repo. 

[...]

PR #338, opened 2017-10-18
update folder repo

danfrankj:
Description of changeset: 
Update the folder repository so that 
1) it doesn't break when getting arguments like `branch` from upstream code 
2) Can init into already created folders
3) Writes references into the folder 

Test Plan: 
```
mkdir test_folder_repo
knowledge_repo --repo test_folder_repo init 
knowledge_repo --repo test_folder_repo create ipynb example.ipynb
knowledge_repo --repo test_folder_repo add example.ipynb -p 'projects/example'
knowledge_repo --repo test_folder_repo runserver
```

Begins to address https://github.com/airbnb/knowledge-repo/issues/332

Auto-reviewers:  @matthewwardrop

### DIFF
--- a/knowledge_repo/repositories/folder.py
+++ b/knowledge_repo/repositories/folder.py
@@ -24,8 +24,15 @@ class FolderKnowledgeRepository(KnowledgeRepository):
     _registry_keys = ['']
 
     @classmethod
-    def create(cls, uri):
-        raise NotImplementedError("Folder does not exist. Not attempting to create it.")
+    def create(cls, uri, embed_tooling=False):
+        path = os.path.abspath(uri)
+        if not os.path.exists(path):
+            raise Exception("Folder does not exist. Not attempting to create it.")
+        shutil.copy(os.path.join(os.path.dirname(__file__), '../config_defaults.py'),
+                    os.path.join(path, '.knowledge_repo_config.py'))
+        shutil.copy(os.path.join(os.path.dirname(__file__), '../templates', 'repo_data_readme.md'),
+                    os.path.join(path, 'README.md'))
+        return FolderKnowledgeRepository(path)
 
     def init(self, config='.knowledge_repo_config.py', auto_create=False):
         self.auto_create = auto_create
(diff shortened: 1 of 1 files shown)


knowledge_repo/repositories/folder.py
```diff
@@ -137,18 +144,12 @@ def _kp_get_revisions(self, path):  # slow
...
-        path = os.path.join(self.path, path)
-        if os.path.isdir(path):
-            ref_path = os.path.join(path, reference)
-            ref_dir = os.path.dirname(ref_path)
-            if not os.path.exists(ref_dir):
-                os.makedirs(ref_dir)
-            with open(ref_path, 'wb') as f:
-                return f.write(data)
-        else:
-            kp = KnowledgePost.from_file(path, format='kp')
```
matthewwardrop: Don't delete this! This adds support for the new standalone KnowledgePost format (a compressed zip of the knowledge post folder that can be shared around).

knowledge_repo/repositories/folder.py
```diff
@@ -137,18 +144,17 @@ def _kp_get_revisions(self, path):  # slow
         raise NotImplementedError()
 
     def _kp_write_ref(self, path, reference, data, uuid=None, revision=None):
-        path = os.path.join(self.path, path)
-        if os.path.isdir(path):
-            ref_path = os.path.join(path, reference)
+        if os.path.isfile(path):
```
danfrankj: @matthewwardrop I put back PKP support PTAL

matthewwardrop: It's not clear to me why checking whether the path is a file is better than checking whether it is a directory, but I have no objection.

You still need the `path = os.path.join(self.path, path)` line, however, otherwise you will be looking for files relative to the current working directory, rather than the repository. `ref_path` should then be returned to just being the join of `path` and `reference`.

danfrankj: Here was the issue... 

When running 
```
knowledge_repo --repo test_folder_repo add example.ipynb -p 'projects/example.kp'
```
several references are written to that path.

 For the first reference, path `projects/example.kp` is not a directory (does not exist) and we therefore do right thing and create that directory and write the reference. Then for the next reference the path IS a directory and so we then try to read it as a PKP. Make sense?

matthewwardrop: Yep... That makes a lot of sense :). Once you've sorted out path issue, and then merge it in :).

PR #342, opened 2017-10-21
Remove catch-all exception handling when verifying posts

jdavidheiser:
Description of changeset: Stop silently suppressing all exceptions

Test Plan: Ran with mix of bystrings and unicodes and confirmed that the errors related to encoding.py actually fired as expected depending on what data was passed around.

### DIFF
--- a/knowledge_repo/post.py
+++ b/knowledge_repo/post.py
@@ -338,7 +338,7 @@ class KnowledgePost(object):
             return False
         try:
             FormatChecks.process(self)
-        except:
+        except AssertionError:
             return False
         return True
 


matthewwardrop:
@jdavidheiser This method is called by the indexer to determine whether a post is valid or not; and if there are encoding errors, the post should not be valid [though the encoding errors are probably a bug we should fix elsewhere]. The idea of this function is see whether we can use this knowledge post without worrying about errors and exceptions, so having it not be a catch-all probably breaks this.

Can you describe the contexts in which you would want this to break at runtime?

jdavidheiser:
It was not an encoding error, it was a "method not found" error switching from Python 2 to 3 - so yeah it was a bug elsewhere, and the generic exception handling obscured the bug.  If there are other Exception types that should be handled, they should be assert statements in FormatCheck, or called out explicitly here (ie UnicodeError)

matthewwardrop:
Agreed. It should definitely break at runtime if there's a bug in the code (at the very least in debug mode). I think the trouble hitherto is that we haven't been particularly disciplined in the types of errors we've thrown around the place. This should most definitely be remedied.

Offhand, I don't know of any other error classes beyond `AssertionError` which should be caught, but I want to check that before merging this... lest we introduce unexpected crashes in the indexer. I'll try to do that soon.

Thanks for the patch!

jdavidheiser:
I'm going to close this PR because there are merge conflicts and it seems like there was not much interest around this cleanup.  I still think this is generally a good idea.

matthewwardrop:
@jdavidheiser Aye... this is definitely the right thing to do... I just never found the time to circle back around to this PR, and do the grunt work of looking for all the different kinds of errors that can be raised intentionally, and then perhaps re-class them so we can explicitly catch those errors.

PR #345, opened 2017-10-27
Add helper flags for init command

aaronbiller:
Description of changeset:
This PR adds 2 flags to the `init` command in the `knowledge_repo` script.
- `--no-prompt`: allows the `init` command to be run without any input from the user (_a la_ Django's `collectstatic --no-input`).
- `--no-config`: a companion to the above flag that when initializing a repo will skip overwriting the `.knowledge_repo_config.py` file.

**Usage**
These changes can be helpful for use in programmatic (ex: upstart) setup of the knowledge repo where the code copied to the server does not include `.git`.
```
$ knowledge_repo --repo ~/krtest init --no-prompt --no-config
WARNING:knowledge_repo.repositories.gitrepository:Overriding existing repository!
Knowledge repository created for uri `/Users/aaronbiller/krtest`.
```


matthewwardrop:
Thanks for this PR too @aaronbiller ! With #386 , I don't think this is needed any more, so I'll close this one out. In particular, #386 makes repository initialisations non-destructive and non-interactive at all times, so these flags probably aren't needed.

Feel free to reopen this PR if you feel I've missed something!

PR #369, opened 2018-01-04
Update server config

aaronbiller:
Description of changeset: 

This fixes the (ironically) explicitly stated caveat of setting the index interval greater than the index timeout.  Additionally, the `server_config.py` file under `resources` is updated to match the current set of config variables under `app.config_defaults.py`.

Test Plan: 
None


matthewwardrop:
Thanks for submitting this @aaronbiller ! It's nice to be configurations with the files in the the repository. Note that the documentation here was in err, rather than the code (which I fixed as part of this PR) [TIMEOUT should be larger than INTERVAL, otherwise every thread will try to claim primary control of the indexing process].

PR #374, opened 2018-01-19
Enable generic OAuth2 config and support self-signed provider CAs

zerogjoe:
The purpose of this PR is to simplify the process of supporting a generic OAuth2 authentication provider when deploying the software, and also to support self-signed CAs for those authentication providers.

Description of changeset: 

- Added a `verify_https` parameter to the `OAuth2Provider` class, whose only purpose is to be passed to the Requests module as the `verify` parameter. This is to support custom CAs (or disabling HTTPS verification entirely)
- Added a generic empty `OAuth2Provider` preset called `oauth2`, the purpose of which is to be overridden by the server administrator in server_config.py.
- Added examples of the use of both of these to `config_defaults.py`
- Synced `resources/server_config.py` with `config_defaults.py` as this is the config template referred to in the README.md, and looks out of date. (Perhaps a better solution is needed to keep these in sync long term.)

Functionality has been tested in production with a self-managed instance of GitLab CE.

### DIFF
--- a/knowledge_repo/app/auth_providers/oauth2.py
+++ b/knowledge_repo/app/auth_providers/oauth2.py
@@ -77,6 +77,7 @@ class OAuth2Provider(KnowledgeAuthProvider):
              client_id=None,
              client_secret=None,
              user_info_mapping=None,
+             verify_https=None,
              validate=None):
 
         (self.base_url,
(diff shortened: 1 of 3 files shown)


aaronbiller:
Hey @zerogjoe, just a heads up that I have a PR #369 open for updating the `server_config.py` file as well.

PR #388, opened 2018-03-26
Rewrote rmd.py converter to add plotly divs and js, and to add plotly…

dorianbrown:
….js dependency

Description of changeset: 

In order to handle plotly plots, we use the rmarkdown::render function, which is a bit more capable than knitr. We render the Rmd to html format, with the option to keep the intermediate .md file. This is almost the same as the knitr output, but contains divs and a json description of the plots.

Additionally some javascript is necessary to have the actual plots rendered. I have a somewhat lazy check for whether we need to add the dependency. This could be improved on by check each code block for either `library(plotly)` or `plotly::`.

I also introduced a new dependency (`python-frontmatter`) to add the javascript tags after the frontmatter. There might be another way of doing this where this additional package isn't needed.

Test Plan: 
A new post is included with three different plotly plots.

### DIFF
--- a/knowledge_repo/converters/rmd.py
+++ b/knowledge_repo/converters/rmd.py
@@ -2,11 +2,27 @@ import os
 import logging
 import subprocess
 import tempfile
+import frontmatter
 
 from ..converter import KnowledgePostConverter
 
 
 logger = logging.getLogger(__name__)
+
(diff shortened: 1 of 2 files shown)


dorianbrown:
It's failing on the additional python-frontmatter dependency.

PR #393, opened 2018-04-01
 Rewrite to rmd.py converter to add plotly divs and js, and to add plotly js dependency

### LINKED ISSUE #387: Plotly in Rmd notebooks being published as images, not javascript

Hi guys,

I've been working on deploying a knowledge repo at my company, and since we're big R users we quickly ran into the issue that Rmd notebooks with Plotly doesn't work as expected. Since the converter uses knitr:
```r
"knit('{fname}', '{target_path}', quiet=F)
```
this tries to convert any plotly plots to static images. I think embedding the javascript plots would be preferable.

I've looked into how we could do this, and I've got some working code ready. I'll try and share it this week, but I've this would be my first contribution to a github project, so I'll need to figure that out first :grin:

**EDIT**: From what I've read, it seems I would need to be added as a contributor to the project in order to push my branch. Would that be possible?

dorianbrown:
Fixes #387.
Description of changeset:

In order to handle plotly plots, we use the rmarkdown::render function, which is a bit more capable than knitr. We render the Rmd to html format, with the option to keep the intermediate .md file. This is almost the same as the knitr output, but contains divs and a json description of the plots.

Additionally some javascript is necessary to have the actual plots rendered. I have a somewhat lazy check for whether we need to add the dependency. This could be improved on by check each code block for either library(plotly) or plotly::.

I also introduced a new dependency (python-frontmatter) to add the javascript tags after the frontmatter. There might be another way of doing this where this additional package isn't needed.

Test Plan:
A new post is included with three different plotly plots.


matthewwardrop:
Hi @dorianbrown , please see my PR into your remote branch here: https://github.com/dorianbrown/knowledge-repo/pull/1 . Otherwise, once that is merged, I think this will be in good shape to land in master.

dorianbrown:
Thanks for the commit, I hadn't realised that Travis was voor Linux and appveyor for windows. I've merged it with mine and it looks the tests are checking it now.

dorianbrown:
`'conda' is not recognized as an internal or external command`

Looks like there's something weong with the python34 windows test setup.

matthewwardrop:
Hi @dorianbrown! Sorry for the hold up, and incommunicado. In the mean time, I've transitioned to a new company, and had my head down ramping up over there. I'm still maintainer of the Knowledge repo, so that's no excuse, but is perhaps understandable.

Less forgivable perhaps is my silence on some technical issues with the PR. In particular, the new method byte64 encodes images into the resulting output, which makes it harder to separate the images from the text, which is necessary to prevent got knowledge repositories becoming prohibitively large over time. I think the right thing to do is probably just extend the image extractor to deal with byte64 encoded images, and after that this can merged in. I'll try to handle that soon.

dorianbrown:
No worries @matthewwardrop, I started a new job last month so I can understand needing some time to focus. Also nice to hear you will still be maintaining the project, as it seems to be getting some momentum.

I can understand your concern for quality solutions, and it sounds like you have a solution in mind. Although I haven't looked at the image extractor code yet, I'd be happy to spend a little time looking into it if you want.

Always fun to learn something new :grin:

matthewwardrop:
@dorianbrown Feel free to look into things if you have some spare time. The code that handles image extraction during the conversion process is here: https://github.com/airbnb/knowledge-repo/blob/master/knowledge_repo/postprocessors/extract_images.py . Quickly looking over it, I'm not sure what this does with byte64 encoded images. Check it out and let me know. Adding support for extracting byte64 encode uris shouldn't be too tricky, if you want to take a stab at it.

tests/test_posts/sample.rmd
```diff
@@ -0,0 +1,39 @@
...
+```{r cars}
+summary(cars)
+```
+
+## Including Plots
+
+You can also embed plots, for example:
+
+```{r pressure, echo=FALSE}
+plot(pressure)
```
matthewwardrop: Do we need a `library(plotly)` or some such here?

tests/test_posts/sample.rmd
```diff
@@ -0,0 +1,39 @@
+---
```
matthewwardrop: Can we rename this file: plotly.rmd (so we know why the file exists).

matthewwardrop:
@dorianbrown Okay. I sit corrected. I think this new rendering method does output things correctly and in a way that the existing parsers understand. Great! Thanks for looking into that.

While it is unfortunate that we need to special case support for Plotly in Rmd files (as compared to Python where the required HTML is automatically added to the notebook), this does seem to be an improvement (and Google searches about other approaches have not yielded any positive results). As such, this is probably good to merge in.

I added a few last comments to the PR around the provided test file. It doesn't seem to explicitly use plotly. Is that expected? It might be good to change the check for whether the header be added to look for the 'library(plotly)' string, if you think that makes sense.

Once we've sorted out these small details, I'll merge it in :). Thanks again for working on this!

matthewwardrop:
@dorianbrown Bump? Will be releasing a new version very soon, and would be nice to get this in :).

MeganBeckett:
This would be a wonderful addition!

PR #396, opened 2018-04-11
Add more complete support for header auth.

matthewwardrop:
This PR adds support for setting user attributes using auth headers.

Build errors are due to a fussier pycodestyle 2.4.0 . Will fix in a separate push.

### DIFF
--- a/knowledge_repo/app/app.py
+++ b/knowledge_repo/app/app.py
@@ -101,8 +101,22 @@ class KnowledgeFlask(Flask):
         def load_user(user_id):
             return User(identifier=user_id)
 
-        # Attempt login via http headers if header is specified
-        if self.config.get('AUTH_USER_IDENTIFIER_REQUEST_HEADER'):
+        # Attempt login via http headers
+        if self.config.get('AUTH_USE_REQUEST_HEADERS'):
+            @self.login_manager.request_loader
+            def load_user_from_request(request):
+                user_attributes = current_app.config.get('AUTH_MAP_REQUEST_HEADERS')(request.headers)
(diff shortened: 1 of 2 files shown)


knowledge_repo/app/app.py
```diff
@@ -101,8 +101,22 @@ def __init__(self, repo, db_uri=None, debug=None, config=None, **kwargs):
         def load_user(user_id):
             return User(identifier=user_id)
 
-        # Attempt login via http headers if header is specified
-        if self.config.get('AUTH_USER_IDENTIFIER_REQUEST_HEADER'):
+        # Attempt login via http headers
+        if self.config.get('AUTH_USE_REQUEST_HEADERS'):
+            @self.login_manager.request_loader
+            def load_user_from_request(request):
+                user_attributes = current_app.config.get('AUTH_MAP_REQUEST_HEADERS')(request.headers)
```
danfrankj: asking to learn... 

would it be possible to bypass this authorization by brute forcing information into your headers?  Or is the idea that you'd only use this method if KR was unreachable except via the proxy which has its own auth?

matthewwardrop: You could in-principle inject values into the headers by subclassing the KnowledgeFlask and playing with the request object, to directly inject LDAP auth or something like that; but then you'd be better off proxying it and handling the auth outside of Knowledge Repo, or setting up an LDAP `AuthProvider`. This PR caters to the use case that the proxy server has added information about the user to the request headers beyond just their username (such as email, photo, etc). So yes... generally this method will only be used if the only route to the Knowledge Repo server is via an authenticating proxy server.

PR #407, opened 2018-05-03
Avoid multiple slashes in oauth redirect_uri

tuvistavie:
Description of changeset: 
I was getting a redirect_uri with to many slashes:

```
http://sub.example.com///oauth/....
```

This should fix the issue.

Test Plan: 

I could not find tests for `OAuthProvider`, but I will try to work on that when I have time.

### DIFF
--- a/knowledge_repo/app/auth_providers/oauth2.py
+++ b/knowledge_repo/app/auth_providers/oauth2.py
@@ -1,3 +1,4 @@
+import posixpath
 import json
 from flask import request, redirect
 from six.moves.urllib.parse import urljoin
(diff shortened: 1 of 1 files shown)


knowledge_repo/app/auth_providers/oauth2.py
```diff
@@ -112,9 +113,13 @@ def init(self,
...
             self.verify_ssl_certs = True
 
-        redirect_url = self.app.config['SERVER_NAME'] or 'localhost:7000'
+        host = self.app.config['SERVER_NAME'] or 'localhost:7000'
+        scheme = self.app.config['PREFERRED_URL_SCHEME'] or 'http'
+        redirect_url = '{}://{}'.format(scheme, host)
+        redirect_path = '/auth/login/{}/authorize'.format(self.name)
         if self.app.config['APPLICATION_ROOT']:
-            redirect_url = '/'.join((redirect_url, self.app.config['APPLICATION_ROOT']))
+            redirect_path = path.join(self.app.config['APPLICATION_ROOT'], redirect_path)
```
matthewwardrop: Using `path.join` will break this on Windows machines (since the default path separator the is '\'). Consider using `'\'.join()` or if you need `path` for some reason, `posixpath`.

tuvistavie: Thank you, I completely forgot about Windows :sweat: 
`join` is needed, as the default `APPLICATION_ROOT` is `/`, which would create an invalid path.

tuvistavie:
Thank you for the review, this is now fixed.

matthewwardrop:
Thanks again for the patch @tuvistavie !

PR #416, opened 2018-05-15
Added LDAP to auth_providers

dorianbrown:
**Description of changeset:**

Implements a simple auth blueprint for LDAP user verification. Currently uses username entered into the HTML from as the uid/cn for verification. I'm only familiar with how my organization has implemented this, but others can implement a username parsing function if they want.

Also updated the login page to look a little better, and require a password.

**Test Plan:**

I tested this using [this LDAP docker image](https://github.com/rroemhild/docker-test-openldap), but it would take some work to add this to the travis/appveyor tests. Any ideas on this?


knowledge_repo/app/auth_providers/ldap.py
```diff
@@ -0,0 +1,25 @@
...
+    _registry_keys = ['ldap']
+
+    def init(self):
+        self.server = Server(self.app.config['LDAP_SERVER'], get_info=ALL)
+
+    def prompt(self):
+        return render_template('auth-login-form.html', skip_password=False)
+
+    def validate(self, user):
+        userdn = self.app.config['USERDN_SCHEMA'].format(user.identifier)
```
matthewwardrop: Two small requests:
- Can we change the name of the config key here to `LDAP_USERDN_SCHEMA` (just to make it semantically clear in the config file that these are related)?
- Can we format it using: `.format(user_id=user.identifier)` so that we can use '{user_id}` in the config file instead of `{0}`?

knowledge_repo/app/auth_providers/ldap.py
```diff
@@ -0,0 +1,25 @@
...
+
+from ..models import User
+from ..auth_provider import KnowledgeAuthProvider
+
+
+class LdapAuthProvider(KnowledgeAuthProvider):
+    _registry_keys = ['ldap']
+
+    def init(self):
+        self.server = Server(self.app.config['LDAP_SERVER'], get_info=ALL)
```
matthewwardrop: Can we add a check here to see if LDAP_SERVER is in the config and not `None`, and if so, raise a polite error? [I'll be suggesting we don't set an LDAP_SERVER by default below].

Maybe:
```
if not self.app.config.get('LDAP_SERVER'):
    raise RuntimeErrror("Use of LDAP authentication requires specification of the LDAP_SERVER configuration variable.")
self.server = ...
```

knowledge_repo/app/config_defaults.py
```diff
@@ -171,6 +171,18 @@ def AUTH_USER_ATTRIBUTE_SETTER(user):
...
+# LDAP configuration
+# ---------------------------------------------------
+# When using an LDAP server for user verification, you need to configure
+# the location of the server, and the directory structure used by your
+# organization.
+
+# Currently the port and protocol must both be included in the server address
+LDAP_SERVER = 'ldaps://127.0.0.1:389'
+# When entering this, note the "{0}" which denotes where the user_id is inserted.
+USERDN_SCHEMA = 'cn={0},ou=people,dc=planetexpress,dc=com'
```
matthewwardrop: As suggested above: `# LDAP_USERDN_SCHEMA = 'cn={user_id},ou=people,dc=planetexpress,dc=com`

knowledge_repo/app/config_defaults.py
```diff
@@ -171,6 +171,18 @@ def AUTH_USER_ATTRIBUTE_SETTER(user):
...
 
+# ---------------------------------------------------
+# LDAP configuration
+# ---------------------------------------------------
+# When using an LDAP server for user verification, you need to configure
+# the location of the server, and the directory structure used by your
+# organization.
+
+# Currently the port and protocol must both be included in the server address
+LDAP_SERVER = 'ldaps://127.0.0.1:389'
```
matthewwardrop: Let's comment it out by default so that if not present an appropriate error message is thrown by the LDAP AuthProvider.

knowledge_repo/app/templates/auth-login-form.html
```diff
@@ -6,7 +6,8 @@
 .login-form {
     display: block;
     list-style-type: none;
-    margin: 10px;
+    width: 340px;
```
matthewwardrop: Can we separate the cosmetic changes from the LDAP patch? Or if its a clear improvement, can you provide a screenshot of the improvement?

dorianbrown: Here's a screenshot of the cosmetic changes. I think it's a clear improvement, below you can see a comparison between the two:

New:
screenshot-2018-5-23 knowledge repo login

Old:
screenshot-2018-5-23 knowledge repo login 1

matthewwardrop: This is indeed an improvement! Leave it in as part of this patch :).

dorianbrown:
All the feedback is good and super useful, I've implemented the changes you mentioned. I also did some work on handing back errors to the login page when stuff goes wrong, but I wanted to change the implementation a bit. 

This seems like a good PR to add it, so I'll try and pretty it up.

matthewwardrop:
@dorianbrown Just wanting to check in again and see what the status of this PR was. Is the LDAP support working nicely for you? If so, we can merge it without unittests and loop back around later, if that is helpful (since I think others might find this valuable as is). The changes to the Rmd backend to support plotly should probably be done as a separate PR though.

If this PR is ready, I'll do a final review, make any necessary convention changes, and get it merged in ASAP :).

matthewwardrop:
@dorianbrown Bump? I cannot actually test this locally, not having an LDAP system set up.

ManuelAudibene:
Hello, since I am interested in LDAP, I would try to check the changes and replicate them in our environment and share my feedback :)

PR #417, opened 2018-05-17
__all__ expects a list, not a dict

BioGeek:
Current behaviour:
```

In [1]: from knowledge_repo.repositories import *
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-3-9fcfb5027797> in <module>()
----> 1 from knowledge_repo.repositories import *

TypeError: 'dict' object does not support indexing

```
The reason is that you're [overriding `__all__`](https://github.com/airbnb/knowledge-repo/blob/master/knowledge_repo/repositories/__init__.py#L3), but `__all__`expects a list and [`import_submodules`](https://github.com/airbnb/knowledge-repo/blob/master/knowledge_repo/utils/import_submodules.py) returns a `dict`.

Description of changeset: 

The code from `import_submodules` is probably taken from [this Stackoverflow answer](https://stackoverflow.com/a/25083161/50065). Note that the additional `.keys()` is present in the usage example. I've wrapped that in an additional `list()` because on Python 3 `.keys()` returns a `dict_keys`  object which doesn't support indexing

Test Plan: 

The following now works:

```
In [1]: from knowledge_repo.repositories import *

In [2]: (gitrepository, dbrepository)
Out[3]: 
(<module 'knowledge_repo.repositories.gitrepository' from '/home/BioGeek/Code/knowledge-repo/knowledge_repo/repositories/gitrepository.py'>,
 <module 'knowledge_repo.repositories.dbrepository' from '/home/BioGeek/Code/knowledge-repo/knowledge_repo/repositories/dbrepository.py'>)

```

### DIFF
--- a/knowledge_repo/app/auth_providers/__init__.py
+++ b/knowledge_repo/app/auth_providers/__init__.py
@@ -1,3 +1,3 @@
 from ...utils.import_submodules import import_submodules
 
-__all__ = import_submodules(__name__)
+__all__ = list(import_submodules(__name__).keys())
--- a/knowledge_repo/app/auth_providers/__init__.py
+++ b/knowledge_repo/app/auth_providers/__init__.py
@@ -1,3 +1,3 @@
 from ...utils.import_submodules import import_submodules
 
-__all__ = list(import_submodules(__name__).keys())
+__all__ = import_submodules(__name__)
--- a/knowledge_repo/converters/__init__.py
+++ b/knowledge_repo/converters/__init__.py
@@ -1,3 +1,3 @@
 from ..utils.import_submodules import import_submodules
 
-__all__ = list(import_submodules(__name__).keys())
+__all__ = import_submodules(__name__)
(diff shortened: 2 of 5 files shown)


BioGeek:
Since it seems that you're not doing anything with the dictionary returned by `import_submodules` anyway, it might be cleaner to rewrite `import_submodules` to something like:
```

def import_submodules(package_name):
    """ Import all submodules of a module, recursively

    :param package_name: Package name
    :type package_name: str
    :rtype: list
    """
    package = sys.modules[package_name]
    submodules = []
    for _, name, _ in pkgutil.walk_packages(package.__path__):
        importlib.import_module(package_name + '.' + name)
        submodules.append(name)
    return submodules

```
instead of adding `list(...)` and `.keys()` to all `__init__.py` files.

matthewwardrop:
Hi @BioGeek ! Good catch! I'd gladly accept a patch for the latter, which as you pointed out, probably makes more sense!

PR #419, opened 2018-05-17
Fix: IE compatability mode

dorianbrown:
Description of changeset: 

Currently IE browsers running in compatability mode cause problems with CSS (see [this SO discussion](https://stackoverflow.com/questions/6771258/what-does-meta-http-equiv-x-ua-compatible-content-ie-edge-do)). Unfortunately a lot of companies still run older version of IE with this enabled by default. 

This line of code prevents IE browsers from using compatibility mode when rendering the web app, and fixes any rendering issues.

Test Plan: 


matthewwardrop:
Thanks @dorianbrown! Anyone using the ancient IE browsers will no doubt thank you kindly!

PR #420, opened 2018-05-20
Deploy with HTTPS for uwsgi and gunicorn

dorianbrown:
**Description of changeset:**

In order to be able to deploy using HTTPS, I added an https flag and paths for cert/key files in the server config file. The uwsgi/gunicorn engines now look at the flag, and if true adds to necessary options to deploy with HTTPS.

Additionally `mimetypes.add_type('image/svg+xml', '.svg')` was added to fix an issue with svg not having the correct MIME type (octet-stream) when using https.

**Test Plan:**
At the moment the `run_tests.sh` doesn't contain any tests for deployment. Maybe something we should add?


matthewwardrop:
Thanks so much for this patch @dorianbrown ! Like your previous patches, this one is awesome and valuable. It does require some testing to make sure that everything is working well, and I am desperately short of time at the moment. I think things should lighten up in the next few weeks, and I'll circle back around to this. And yes.. having unittests for this would ease the manual testing burden substantially... if you have some time to write some, and can think of a way to do it sensibly, that would be amazing!

Issue #170, opened 2016-12-12
[Feature] Have a better nesting structure for /cluster view


NiharikaRay:
Folders within folders should be nested. 

Auto-reviewers: @NiharikaRay @matthewwardrop @earthmancash @danfrankj

tcbegley:
This would be an awesome feature, is there any update on whether it's on the roadmap?

dorianbrown:
Sounds pretty doable with a little html/css/js knowledge. 

This issue isn't my highest priority (not the maintainer btw), but feel free to make some changes and create a pull request! Otherwise I might have a look at it in a few months, it does seem pretty basic right now.

tcbegley:
No worries, thanks for the clarification.

I'll take a look and create a pull request if I can figure it out.

dorianbrown:
If we want some pointers for where to look and what to change, let me know and I can help you out. Having a little Flask and Jinja templating knowledge is probably helpful, but its really easy to learn.

tcbegley:
Sure, happy to take any pointers. My current thinking was to modify `render_cluster` in `app/routes/index.py` to create the nested structure, then also modify `app/templates/index-cluster.html` to recursively unpack it and generate the nested lists. I haven't figured out which bits of the css/js will need modifying yet if any. Does that seem like a reasonable starting point?

dorianbrown:
To be honest, I think this can be done using only jinja templating. It's the first time I'm looking at this code, and I haven't done anything too fancy with Jinja, but it seems to be possible to do recursion with just jinja. That would mean you won't have to touch any javascript, and maybe just some CSS to make it look a little prettier. 

Here's a [bit from the documentation](http://jinja.pocoo.org/docs/2.10/templates/):

> It is also possible to use loops recursively. This is useful if you are dealing with recursive data such as sitemaps or RDFa. To use loops recursively, you basically have to add the recursive modifier to the loop definition and call the loop variable with the new iterable where you want to recurse.
> 
> The following example implements a sitemap with recursive loops:
> 
> ```html
> <ul class="sitemap">
> {%- for item in sitemap recursive %}
>     <li><a href="{{ item.href|e }}">{{ item.title }}</a>
>     {%- if item.children -%}
>         <ul class="submenu">{{ loop(item.children) }}</ul>
>     {%- endif %}</li>
> {%- endfor %}
> </ul>
> ```
> 
[...]

PR #424, opened 2018-05-29
Nested folder structure for cluster view

tcbegley:
Description of changeset: 

This PR adds a nested structure to the /cluster view, allowing for more compact categorization of posts. It is intended to resolve issue #170 

Sample screenshot:

image

Very happy to take suggestions on how to properly test this. Also feedback on the javascript changes are particularly welcome as I'm a bit of a novice.

### DIFF
--- a/knowledge_repo/app/routes/index.py
+++ b/knowledge_repo/app/routes/index.py
@@ -7,7 +7,6 @@ This includes:
   - /favorites
 """
 import os
-import posixpath
 import json
 from builtins import str
 from flask import request, render_template, redirect, Blueprint, current_app, make_response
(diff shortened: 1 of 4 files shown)


tcbegley:
Thanks, happy to be able to contribute.

Looking at `run_tests.sh` it looks like the tests create some posts with paths like `projects/test/ipynb_test` for example, so I think we should be covered. Unless we want to add some tests to cover whether the sorting of subdirectories is working correctly for example?

knowledge_repo/app/routes/index.py
```diff
@@ -137,12 +136,14 @@ def render_cluster():
         author_to_posts = {}
         authors = (db_session.query(User).all())
         for author in authors:
-            author_posts = [post for
-                            post in author.posts
-                            if post.is_published and not post.contains_excluded_tag]
+            author_posts = [
+                [post.title, True, 0, post]
```
matthewwardrop: For clarity, can you convert this to use a NamedTuple instance?

Perhaps something like: `ClusterEntry = NamedTuple('name', 'is_post', 'children_count', 'content')`

tcbegley:
Hey @matthewwardrop, thanks for the feedback. I think using `namedtuple` is a good idea, it adds a lot of clarity.

One complication it introduces is in the sorting, since named tuples are immutable and I was relying on reassigning the contents to the return value of the recursive sorting function. I have pushed a hacky fix, that sorts the contents attribute of each tuple in place, however I'm a bit uneasy about it.

Depending on your point of view, I propose using either dictionaries with informative key names, or a custom class as a proxy for a mutable namedtuple instead. This will still add clarity to the code, but won't necessitate hacking the sorting function. Let me know what you think or if you have a preference for either of the two options.

matthewwardrop:
@tcbegley : Goodness.... sorry! Your fixes made the code much more readable, and having gone through the code and tested it locally, I think we're ready to merge it in. Many thanks again for the contribution!

PR #436, opened 2018-06-21
[gunicorn] Adding statsD options

john-bodley:
Description of changeset: 

This PR enables instrumentation to the Gunicorn webserver via the statsD protocol which is configurable via the [statsd_host](http://docs.gunicorn.org/en/stable/settings.html#statsd-host) and [statsd_prefix](http://docs.gunicorn.org/en/stable/settings.html#statsd-prefix) settings.

Note ideally it would  be great to have a mechanism of parsing arbitrary Gunicorn command line arguments which don't need to be individually registered or use a Gunicorn config file for configuring the Gunicorn webserver.

Test Plan: 

Verified that 
```
knowledge_repo deploy
knowledge_repo deploy --statsd-host=localhost:8125
knowledge_repo deploy --statsd-host=localhost:8125 --statsd-prefix=knowledge
```
correctly launched the Gunicorn webserver without error. 

### DIFF
--- a/knowledge_repo/app/deploy/common.py
+++ b/knowledge_repo/app/deploy/common.py
@@ -22,13 +22,17 @@ class KnowledgeDeployer(with_metaclass(SubclassRegisteringABCMeta, object)):
                  host='0.0.0.0',
                  port=7000,
                  workers=4,
-                 timeout=60):
+                 timeout=60,
+                 statsd_host=None,
+                 statsd_prefix=None):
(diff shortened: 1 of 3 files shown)


john-bodley:
@danfrankj @matthewwardrop let me know if you would prefer me to simply add the [config](http://docs.gunicorn.org/en/stable/settings.html#config) file option which provides more flexibility.

PR #437, opened 2018-06-22
[gunicorn] Adding environment configuration

john-bodley:
Description of changeset: 

This PR mimics logic [here](https://github.com/benoitc/gunicorn/blob/master/gunicorn/app/base.py#L150) to allow additional Gunicorn configuration to be defined via the `GUNICORN_CMD_ARGS` environment variable, i.e., 
```
GUNICORN_CMD_ARGS='--statsd-host=localhost:8125' knowledge_repo deploy
```

Test Plan: 

Verified that the command line args where present.

### DIFF
--- a/knowledge_repo/app/deploy/gunicorn.py
+++ b/knowledge_repo/app/deploy/gunicorn.py
@@ -21,12 +21,21 @@ class GunicornDeployer(BaseApplication, KnowledgeDeployer):
         BaseApplication.__init__(self)
 
     def load_config(self):
-        options = {
+        env_args = self.cfg.parser().parse_args(self.cfg.get_cmd_args_from_env())
+
+        # Load up environment configuration.
+        for key, value in vars(env_args).items():
+            if key != 'args' and value is not None:
+                self.cfg.set(key, value)
+
+        # Lastly, update the configuration with any command line settings.
+        command_line_args = {
             'bind': u'{}:{}'.format(self.host, self.port),
             'workers': self.workers,
-            'timeout': self.timeout
+            'timeout': self.timeout,
         }
-        for key, value in options.items():
+
+        for key, value in command_line_args.items():
             self.cfg.set(key, value)
 
     def load(self):
--- a/knowledge_repo/app/deploy/gunicorn.py
+++ b/knowledge_repo/app/deploy/gunicorn.py
@@ -28,14 +28,14 @@ class GunicornDeployer(BaseApplication, KnowledgeDeployer):
             if key != 'args' and value is not None:
                 self.cfg.set(key, value)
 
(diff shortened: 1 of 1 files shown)


knowledge_repo/app/deploy/gunicorn.py
```diff
@@ -21,12 +21,21 @@ def __init__(self, *args, **kwargs):
...
+        # Lastly, update the configuration with any command line settings.
+        args = {
             'bind': u'{}:{}'.format(self.host, self.port),
             'workers': self.workers,
-            'timeout': self.timeout
+            'timeout': self.timeout,
         }
-        for key, value in options.items():
+
+        for key, value in args.items():
```
NiharikaRay: do you still need this iteration over `args.items()`, when you're doing the same thing above on line 26?

knowledge_repo/app/deploy/gunicorn.py
```diff
@@ -21,12 +21,21 @@ def __init__(self, *args, **kwargs):
         BaseApplication.__init__(self)
 
     def load_config(self):
-        options = {
+        args = self.cfg.parser().parse_args(self.cfg.get_cmd_args_from_env())
+
+        # Load up environment configuration.
+        for key, value in vars(args).items():
+            if key != 'args' and value is not None:
```
NiharikaRay: why are we filtering out for keys that are `args` here?

knowledge_repo/app/deploy/gunicorn.py
john-bodley: The args here are different. These are the args from the `GUNICORN_CMD_ARGS ` environment variable.

knowledge_repo/app/deploy/gunicorn.py
john-bodley: `args` here is a config object which contains the `arg` field which should be ignore. This logic is from [here](https://github.com/benoitc/gunicorn/blob/master/gunicorn/app/base.py#L150).

knowledge_repo/app/deploy/gunicorn.py
```diff
@@ -21,12 +21,21 @@ def __init__(self, *args, **kwargs):
...
+                self.cfg.set(key, value)
+
+        # Lastly, update the configuration with any command line settings.
+        args = {
```
danfrankj: @john-bodley for readability and future-proofing, let's have only use `args` once. e.g. above `env_args` and here `command_line_args`

john-bodley: 👍

john-bodley:
@danfrankj I addressed your comment.

Issue #430, opened 2018-06-09
Posts don't show and can't be accessed when using MySQL DB


CPapadim:
Auto-reviewers: @NiharikaRay @matthewwardrop @earthmancash @danfrankj

I'm trying to deploy knowledge repo with a database backend.  The database is running on a different place than the knowledge_repo server and is accessed over an SSH tunnel.

It seems that Knowledge Repo can read from and write to this database just fine.

However, for whatever reason, no posts are showing up.  And it's not just in the feed, but they also can't be accessed directly via their path.  I looked in the database itself and the posts are there in the repository table, however, the `posts` table that knowledge repo makes itself is empty.   If I try to re-add the posts I see in the repository table, I expectedly get the error that their paths already exist.

I tried also setting SQLALCHEMY_DATABASE_URI to use sqlite, but that didn't seem to help.  

Another thing I tried was running a MySQL DB on a local docker container, and the same issue persists.  

Version 0.8.1 of Knowledge Repo, on a Mac, with python 3 if that helps.

I'm sort of at a loss as to why this could be happening.

In the logs, I see stuff like this:

```  File "/usr/local/lib/python3.6/site-packages/MySQLdb/connections.py", line 277, in query
    _mysql.connection.query(self, query)
sqlalchemy.exc.OperationalError: (_mysql_exceptions.OperationalError) (2013, 'Lost connection to MySQL server during query') [SQL: 'SELECT index_metadata.id AS index_metadata_id, index_metadata.type AS index_metadata_type, index_metadata.name AS index_metadata_name, index_metadata.value AS index_metadata_value, index_metadata.updated_at AS index_metadata_updated_at \nFROM index_metadata \nWHERE index_metadata.type = %s AND index_metadata.name = %s \n LIMIT %s'] [parameters: ('lock', 'master_check', 1)] (Background on this error at: http://sqlalche.me/e/e3q8)
```

During startup of knowledge repo I also see this:

```WARNING:knowledge_repo.app.index:Master indexing thread has died. Restarting...```

CPapadim:
It looks like although the Lost Connection errors with `runserver` go away, they come back when I use `deploy`.  

I tried switching to postgres DB and I get lost connection errors with that too `server closed the connection unexpectedly`

It's strange that this is only happening with the indexing query and other queries seem to be able to read and write into the DB.

irdvl:
👍 - I have a similar problem when running `gunicorn` driven Knowledge Repo server against a Postgres instance.

[...]

PR #445, opened 2018-08-23
Db connection fixes

CPapadim:
Description of changeset: 

Fixes to indexing, including properly closing connections and writing repo revision to the DB

See Issue #430 

Test Plan: 


matthewwardrop:
Thanks for spending so much time tracking these heisenbugs down. I really appreciate it (especially since I haven't had the time to do so myself). Since these changes only affect indexing, and not runtime performance of queries, the small overhead likely to be incurred here shouldn't matter. Merging it in! Thanks again @CPapadim !

PR #455, opened 2018-09-27
Revert "Db connection fixes"

matthewwardrop:
Reverts airbnb/knowledge-repo#445 . Looks like I should have reviewed this a bit more closely @CPapadim ! This is causing issues with the in-memory sqlite database connection, because throwing away the engine also throws away the database.

Can you see if you can fix the issue by instead re-establishing the connection from the indexing thread instead of on the model itself? [I think that's how I thought it was being done already, but obviously didn't look closely enough >.<].


matthewwardrop:
Hmmm... I don't want to revert everything. Let's do this a different way.

PR #456, opened 2018-09-27
Configure Travis tests for Python versions

naoyak:
Description of changeset: 
Travis scripts were not actually using the Python versions specified in the build matrix (`conda` was installing 3.7 for the Python 3 build).

Test Plan: 
Run CI

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,9 +7,9 @@ environment:
       PYTHON_ARCH: "32"
       MINICONDA: C:\Miniconda
 
-    - PYTHON: "C:\\Python34"
-      PYTHON_VERSION: "3.4.1"
-      PYTHON_ARCH: "32"
+    - PYTHON: "C:\\Python37-x64"
+      PYTHON_VERSION: "3.7.0"
+      PYTHON_ARCH: "64"
       MINICONDA: C:\Miniconda3
 
 init:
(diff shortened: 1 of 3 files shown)


tests/config_server.py
```diff
@@ -3,7 +3,7 @@
 # This file is only needed for setting up a dedicated server configuration
 # with support for extracting username information and emailing users.
 
-SERVER_NAME = 'localhost'
+SERVER_NAME = 'localhost.localdomain'
```
matthewwardrop: Why this change? I think this will lead to platform specific behaviour, no?

appveyor.yml
```diff
@@ -7,9 +7,9 @@ environment:
       PYTHON_ARCH: "32"
       MINICONDA: C:\Miniconda
 
-    - PYTHON: "C:\\Python34"
-      PYTHON_VERSION: "3.4.1"
-      PYTHON_ARCH: "32"
+    - PYTHON: "C:\\Python37-x64"
+      PYTHON_VERSION: "3.7.0"
```
matthewwardrop: We only explicit claim support for Python 3.6, which may change soon; but in the mean time, let's target 3.6.6 here.

tests/config_server.py
naoyak: @matthewwardrop so I noticed that the test runs have a bunch of [Flask warnings](https://travis-ci.org/airbnb/knowledge-repo/jobs/433841809#L1065-L1109), and this makes them go away. This config is only for running the test suite, right?

matthewwardrop: @naoyak Yep... I was worried it might break Windows... but it didn't! Merging!

PR #460, opened 2018-10-03
Add subtitle field and allow interactive specification of headers.

matthewwardrop:
This patch-set improves the functionality around headers and their presentation.

In particular:
- [x] Adds support for a subtitle field.
- [x] Improves the rendering of headers (and removes the incorrect treatment of document title as a header; freeing up h1 for proper use).
- [x] Improves readability by slightly increase font size and reducing maximum page width.
- [x] Allows for interactive specification of missing but required headers (will be accessible via the new `kp` tooling).
- [x] Slightly improves the styling associated with headers, which formerly could be smaller than the body text.
- [x] Fixes requirement for `updated_at` header to be present even though it is later overwritten.

screenshot from 2018-10-03 13-46-07

### DIFF
--- a/knowledge_repo/_version.py
+++ b/knowledge_repo/_version.py
@@ -26,6 +26,7 @@ __dependencies__ = [
     'gitpython',  # Git abstraction
     'tabulate',  # Rendering information prettily in knowledge_repo script
     'pyyaml',  # Used to configure knowledge repositories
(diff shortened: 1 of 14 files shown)


matthewwardrop:
@danfrankj I've checked this over pretty rigorously locally. Merging in now. Let me know if you see anything concerning retrospectively.

Issue #278, opened 2017-05-30
Google charts - month indexing starts at 0


jeffreygrover:
Hi,

This is a _very_ minor thing, but I've noticed that the 'Stats' page uses Google charts. The months in JavaScript dates are 0-indexed, so all of the plots are off by a month. 

Thanks for creating and maintaining such a great tool!

-Jeff

Auto-reviewers: @NiharikaRay @matthewwardrop @earthmancash @danfrankj

matthewwardrop:
Thanks Jeff. We'll fix this for the next release :).

PR #474, opened 2018-12-13
Subtract 1 from months when making JavaScript Dates for use in Google charts

srowen:
Subtract 1 from months when making JavaScript Dates for use in Google charts. JS months are 0-indexed.

Description of changeset: 

I believe this resolves https://github.com/airbnb/knowledge-repo/issues/278
This seems to be the jinja2 way of subtracting from a variable's value:
http://jinja.pocoo.org/docs/2.10/templates/#math

Test Plan: 

`./run-tests.sh` passes. I admit I'm not sure how to run the locally built copy of knowledge repo to check the change.

### DIFF
--- a/knowledge_repo/app/templates/stats.html
+++ b/knowledge_repo/app/templates/stats.html
@@ -51,7 +51,7 @@ function drawChart() {
   data.addColumn('number', 'Pageviews');
   data.addRows([
           {%- for k, v in daily_pageviews.items() %}
-          [new Date({{ k.year }}, {{ k.month }}, {{ k.day }}), {{ v }}],
+          [new Date({{ k.year }}, {{ k.month - 1 }}, {{ k.day }}), {{ v }}],
           {%- endfor %}
   ]);
   data.sort(0);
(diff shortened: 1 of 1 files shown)


matthewwardrop:
@srowen Thanks for this patch as well! It looks good. Should the day also be zero-indexed?

srowen:
Thanks again for the tip on running locally @matthewwardrop . Yes it seems to fix the display issue.
No only months are 0-indexed: https://www.w3schools.com/js/js_date_methods.asp
This mirrors Java FWIW.

matthewwardrop:
@srowen This patch looks good also. I'll merge it in. Thanks for fixing these papercut issues.

Thanks also for being thorough and for the JS reference. I've done quite a bit of JS work myself in other projects, including work with dates, but always end up having to check the idiosyncrasies of each datetime library lest I make these off-by-one errors (like this one, though at least this one in particular was not of my hand ;).

srowen:
Thanks! just little things indeed but help our internal usage at Databricks, where it's pretty helpful. Maybe I can add our org to INTHEWILD.md

matthewwardrop:
@srowen Feel free to add your org to the INTHEWILD.md file :). Glad it is useful to you!

Issue #363, opened 2017-12-14
Broken Sorting in Table View


danfrankj:
![image](https://user-images.githubusercontent.com/616139/34019844-a2606bc8-e0e5-11e7-8a6e-adcfa202122b.png)

Auto-reviewers: @NiharikaRay @matthewwardrop @earthmancash @danfrankj

PR #475, opened 2018-12-13
Fix sorting by 'views' column in Table view

srowen:
Description of changeset: 

I believe this could resolve https://github.com/airbnb/knowledge-repo/issues/363
The issue is that `index-table.html` uses these column names when forming the `sort_by` query parameter ...
```
var col_names = ['Title', 'Author', 'CreatedAt', 'UpdatedAt', 'Views', 'Upvotes', 'Comments']
```
... but `posts.py` is expecting sort_by columns like ...
```
    post_properties = {
        "updated_at": Post.updated_at,
        "created_at": Post.created_at,
        "title": Post.title,
    }
    join_order_col = {
        "uniqueviews": func.count(distinct(PageView.user_id)),
        "allviews": func.count(PageView.object_id),
        "upvotes": func.count(Vote.object_id),
        "comments": func.count(Comment.post_id)
    }
```

`views` isn't among the options but `uniqueviews` and `allviews` are. I think this should be an alias for `allviews`?

Test Plan: 

`./run_tests.sh` passes, but I am not sure how to run this locally to test the sorting.

### DIFF
--- a/knowledge_repo/app/utils/posts.py
+++ b/knowledge_repo/app/utils/posts.py
@@ -89,8 +89,8 @@ def get_posts(feed_params):
         (join_table, join_on) = joins[sort_by]
 
         query = (db_session.query(Post, order_col)
-                 .outerjoin(join_table, Post.path == join_on))
-        query = query.group_by(Post.path)
+                 .outerjoin(join_table, Post.id == join_on))
+        query = query.group_by(Post.id)
 
     # sort order
     if order_col is not None:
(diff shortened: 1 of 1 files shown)


matthewwardrop:
Thanks for the patch @srowen ! I don't think I've ever touched this piece of code in the knowledge repo since I got involved in the project; I'll take a look soon. From a quick glance, the patch looks good.

Regarding testing it locally, you should be able to run:
`./<repo_path>/scripts/knowledge_repo <knowledge repository git repo or folder> runserver --debug --port 7000`

matthewwardrop:
Whoops! Yes, you need the `--repo` flag. I've modified my comment above to include it for posterity.

I'm happy to have the author sorting fixed separately. I've also tossed around the idea of removing this table view, but only after we have proper sorting and filtering of the main index "feed"... which would probably take more time than I have to dedicate to the project at the moment. In the meantime, fixes like these are really valuable in terms of keep the project usable. So thanks again!

This patch looks good now, so I'll merge it in!

Issue #423, opened 2018-05-29
confusing documentation for database uri


tam-borine:
Auto-reviewers: @NiharikaRay @matthewwardrop @earthmancash @danfrankj
When I tried to use `--repo mysql://username:password@hostname/database:table_name` it threw the error that I didn't have the database. But I had connected to it with psql so that wasn't true. Turned out removing the ":table_name" substring worked for me! Maybe the documentation should be clearer?

PR #476, opened 2018-12-14
Clarify DB URI docs, and fix display of --flags that render as -flags in docs (em-dash)

srowen:
Description of changeset: 

Clarify DB URI docs, and fix display of --flags that render as -flags in docs (em-dash).
I believe this is what https://github.com/airbnb/knowledge-repo/issues/423 was requesting for doc updates. I added a note about sqlite URIs as that worked for us.

Separately I also fixed a small issue, that sphinx seems to collapse `--` to an em-dash for flags like `--dburi` in the docs, which makes them incorrect as `-dburi`. That also tripped me up for a second.

(There are many instances where code-like text is back-ticked and actually renders as italics, but I didn't bother with those.)

Test Plan: 

Manually built docs and verified.

### DIFF
--- a/docs/deployment.rst
+++ b/docs/deployment.rst
@@ -119,9 +119,9 @@ by running:
 which starts a web application instance on `http://127.0.0.1:7000` with the
 default (insecure) options. The command line also supports some high-level
 options, such as :code:`--port` and :code:`--dburi` which respectively change the local
-port on which the server is running, and the SQLAlchemy URI where the database
-can be found and/or initiated (which is necessary even when not using a
-database-backed knowledge repo).
+port on which the server is running, and the SQLAlchemy URI of the database to use
+and/or initialize for the post index (which is distinct from the database, if any,
+used to store knowledge posts).
 
 For shared deployments, however, you will probably need to create a server
 configuration file. A complete server configuration template can be found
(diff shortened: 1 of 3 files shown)


docs/deployment.rst
```diff
@@ -118,9 +118,10 @@ by running:
 
 which starts a web application instance on `http://127.0.0.1:7000` with the
 default (insecure) options. The command line also supports some high-level
-options, such as `--port` and `--dburi` which respectively change the local
-port on which the server is running, and the sqlalchemy uri where the database
-can be found and/or initiated.
+options, such as :code:`--port` and :code:`--dburi` which respectively change the local
+port on which the server is running, and the SQLAlchemy URI where the database
```
matthewwardrop: Perhaps instead of your addendum: "and the SQLAlchemy URI of the database to use and/or initialize for the post index (which is distinct from the database, if any, used to store knowledge posts)."

matthewwardrop:
Thanks again for the nit fixes @srowen !

Are you using the knowledge repo internally somewhere, or just a good samaritan? :).

matthewwardrop:
@srowen Ah! That's interesting. How do you automate the conversion to the Knowledge Repo format?

srowen:
Not much automation really -- Databricks can export notebooks as .ipynb and knowledge repo can import that, and that's more than enough. It can be done via APIs on the command line. Our instance more of a search engine and viewer so if there are any minor issues with markdown in conversion, doesn't matter.

matthewwardrop:
@srowen Thanks again for the patch. I'll merge it in now. Thanks for sharing re:the automation process. Always interesting to hear how others are adopting the code.

Issue #428, opened 2018-06-04
Insistent request of Git credentials


pveragen:
Auto-reviewers: @NiharikaRay @matthewwardrop @earthmancash @danfrankj

Hi, 

I'm testing a knowledge repo implementation for our team using a git backend.  Until know I could manage the deploy and Oauth using github, but I have an annoying problem of insistent and multiple request of my git user and password (no restart involved).  I'm using the flask engine (and sqlite database).  After the deploy the 

`INFO:knowledge_repo.repositories.gitrepository:Fetching updates to the knowledge repository...`

routine request the info. If no provided, the server doesnt load the web 

<img width="1019" alt="screen shot 2018-06-04 at 5 47 01 pm" src="https://user-images.githubusercontent.com/36960548/40943568-9a9b0ea4-681f-11e8-84bf-8779e6b24396.png">

I dont know if I'm missing some config

Thanks,

dorianbrown:
I think setting up your repository using an ssh key would solve this problem.

matthewwardrop:
Thanks for handling this @dorianbrown :). I'll close this issue once it's been documented :).

PR #477, opened 2018-12-14
Add note about setting up Github authentication with ssh

srowen:
Description of changeset: 

Add note about setting up Github authentication with ssh.
This is intended to document the suggestion in https://github.com/airbnb/knowledge-repo/issues/428 and this has worked for me. 

Test Plan: 

Built and checked the docs.

### DIFF
--- a/docs/deployment.rst
+++ b/docs/deployment.rst
@@ -56,26 +56,30 @@ whichever service you plan to use) and then running:
 Users can then clone this repository, and point their local `knowledge_repo`
 script at it using :code:`--repo <path_of_cloned_repository>`.
 
-If the git repo is hosted on a remote service like Github, then the knowledge
-repo server will periodically fetch updates, while running as a server. If
-the remote service requires authentication (e.g. a private repository on Github),
-then these requests will fail.
+When serving the posts from this repository using the Knowledge Repo web application,
+it will attempt to periodically fetch upstream changes to the repository. This can
+be disabled in the server configuration by setting :code:`INDEXING_UPDATES_REPOSITORIES=False`,
+in which case you will have to create your own external process for updating the repository
+in order for the Knowledge Repo to see new posts.
 
-Taking Github as an example, it's possible to use SSH to authenticate in order
-to allow the knowledge repo server access. To set this up:
+If the git repository is served on a remote host that requires authentication, the
+integrated fetching routines will fail unless you have properly set up your environment
+with the appropriate authentication.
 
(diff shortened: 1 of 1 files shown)


docs/deployment.rst
```diff
@@ -56,6 +56,27 @@ whichever service you plan to use) and then running:
 Users can then clone this repository, and point their local `knowledge_repo`
 script at it using :code:`--repo <path_of_cloned_repository>`.
 
+If the git repo is hosted on a remote service like Github, then the knowledge
+repo server will periodically fetch updates, while running as a server. If
+the remote service requires authentication (e.g. a private repository on Github),
+then these requests will fail.
```
matthewwardrop: What do you think of the following rewording of this paragraph:

When serving the posts from this repository using the Knowledge Repo web application, it will attempt to periodically fetch upstream changes to the repository. This can be disabled in the server configuration by setting `INDEXING_UPDATES_REPOSITORIES=False`, in which case you will have to create your own external process for updating the repository in order for the Knowledge Repo to see new posts. If the git repository is served on a remote host that requires authentication, the integrated fetching routines will fail unless you have properly set up your environment with the appropriate authentication.

docs/deployment.rst
```diff
@@ -56,6 +56,27 @@ whichever service you plan to use) and then running:
...
+
+Taking Github as an example, it's possible to use SSH to authenticate in order
+to allow the knowledge repo server access. To set this up:
```
matthewwardrop: Github -> GitHub?

I would recommend that people use deploy keys (which are repository specific) rather than a personal ssh key, and can be granted read-only access. (See: https://developer.github.com/v3/guides/managing-deploy-keys/#deploy-keys).

It's also possible to bypass some of this configuration by preparing the ssh-agent in the environment to be preloaded with a specific key, for example: `ssh-add /decrypted_secrets/id_rsa`, and adding the host to the list of known hosts: `ssh-keyscan -t rsa -p <ssh port> <host> >> ~/.ssh/known_hosts`.

Thoughts?

srowen: Oh that's good to know. Let me try to set that up internally, and then update the guide accordingly.

srowen: Yep @matthewwardrop that's a much better solution and works. I learned something. I updated the guidance accordingly with your other edits.

matthewwardrop:
@srowen Thanks again for your documentation improvements!

PR #488, opened 2019-01-04
Fix typo "nane"->"name"

schillerk:
The "Cluster" button isn't receiving the same margin between lable and icon as "Card" and "Table".

screen shot 2019-01-03 at 8 07 31 pm

Description of changeset: 

Test Plan: 


matthewwardrop:
Thanks for the fix @schillerk. Merging in now.

PR #491, opened 2019-01-17
Docker

perryism:
Description of changeset: 
Make it easy to setup knowledge repo on a local machine.  

Test Plan: 
1. download docker
1. build an image by running `docker build -t knowledge -f docker/Dockerfile.dev .`
1. find a folder where your notebooks are and executing the following command with the folder
1. run the docker image by `docker run --rm -it -p 7000:7000 -v <notebook-folder-path>:/data -e KNOWLEDGE_REPO=test_repo knowledge`  


docker/README.md
```diff
@@ -0,0 +1,15 @@
+# Docker
+
+Make sure you have Docker installed
+
+## Build an image
+
+<pre>
+docker build -t knowledge -f docker/Dockerfile.dev .
```
OFranke: That command looks like you would want to run it from the root directory. With the `.` at the end you are setting the build context to root which will actually result in the build to fail at line 22 `RUN pip install -r requirements.txt`.

The reason is that in line 20 `COPY . /app/knowledge-repo-$VERSION` you are copying the given build context. in that case the whole knowledge-repo to `/app/knowledge-repo-$VERSION`. Your build will fail because requirements will not be found. You expect it to be inside the container in `/app/knowledge-repo-$VERSION/requirements.txt` but it will actually be in `/app/knowledge-repo-$VERSION/docker/requirements.txt`. The same thing would happen for `entrypoint.sh`

To fix that you need to specify proper build context and change the command to `docker build -t knowledge -f docker/Dockerfile.dev ./docker`

docker/Dockerfile.dev
```diff
@@ -0,0 +1,30 @@
...
+    && rm -rf /var/lib/apt/lists/* 
+
+ARG VERSION=0.8.8
+
+RUN wget https://github.com/airbnb/knowledge-repo/archive/v$VERSION.zip && \
+    unzip v$VERSION.zip -d /app
+
+WORKDIR /app/knowledge-repo-$VERSION
+
+COPY . /app/knowledge-repo-$VERSION
```
OFranke: Why do you copy the whole directory into `/app/knowledge-repo-$VERSION`?
Wouldn't it be enough to just copy `requirements.txt`?

docker/Dockerfile.dev
```diff
@@ -0,0 +1,30 @@
...
+
+RUN pip install -r requirements.txt
+
+COPY entrypoint.sh /app/knowledge-repo-$VERSION
```
OFranke: When copying shell scripts you should also set sufficient permissions inside the container. Otherwise the permissions of your host system are used as default.

You could add `RUN chmod +x /app/knowledge-repo-$VERSION/entrypoint.sh` or instead specify a user that is running the process.

omrihar:
Is this a dead PR? Are there any news about creating a nice Docker image for the Knowledge repo? I've struggled quite a lot in just getting it to work on my local machine with the simplest possible setup I could imagine. Just using Docker would be really great!

JavierLopezT:
It would be awesome to have Docker for this project

csharplus (approved):
I've tested this PR out locally, and made some updates to this PR to fix some issues encountered.

PR #502, opened 2019-03-19
Adjust regex matching markdown images to less eagerly match beyond the tag in brackets

srowen:
Description of changeset: 

When adding an ipython notebook, a postprocessor scans for images in the JSON by searching with a regex for markdown-style images like `[tag](url)`. The regex is as you see below. 

I had a cell with some inline Javascript like: `![123]...lots of stuff...](...stuff...)`. This matched as an image because the reluctant quantifier still forces it to expand the match of the tag until it sees "](". It shouldn't match.

The current regex looks for "space, or not space" in between the square braces. That didn't sound right -- why not "."? But in any event my guess is the intent is to match "anything except a closing square brace".

This at least made my case work. Submitted for your consideration.


srowen:
Serendipitously, I just learned on Twitter what the [\s\S] construction is for - matches line breaks unlike '.'. I think the fix is still probably valid? Would you match a tag across newlines?

naoyak:
Sorry @srowen - can you provide a simple example which breaks the parser?

srowen:
Yeah I should have saved that, though it was unfortunately internal. I don't have a repro right now. OK to close until I or someone else does; it hasn't happened recently.

naoyak (approved):

PR #511, opened 2019-04-23
added dedent to text/html in ipynb.py

amniskin:
Without the dedent, libraries that add HTML without dedenting beforehand (which should be unnecessary) tend to show up as verbatim tags (due to markdown's interpretation of indents). So to account for this, we can dedent each text/html node before writing them out.

Description of changeset: 
converters.ipynb

Test Plan: 


amniskin:
Submitted a PR to nbconvert instead. It works for me. Plotly should be working now, ya'll. Yay!

PR #515, opened 2019-05-15
remove auto-reviewers who are not watching anyways

danfrankj:
### Description of changeset: 
Matthew is probably the only one really looking at these anymore so we should just remove this altogether. 

### Test Plan: 
None

Last time...


matthewwardrop:
Let it be done :).

PR #517, opened 2019-06-05
Improve subscription emails

### LINKED ISSUE #516: Fix & improve subscription emails
### Context

When a new post is created within a tag, an email is sent to all users that are subscribed to that tag (if you have this functionality enabled). We have been testing this feature before using it in our production knowledge repo and found some issues.

### Problems

- The whole post is rendered and sent in the email, which is a problem per se if the post is too large. Besides, it is problematic when you have images, graphs and the like. What the receiver sees is a bunch of illegible code.

- Images are "attached" using its base64 representation (``), which is not supported by several email clients (including gmail, check [this](https://user-images.githubusercontent.com/483532/58936053-2f86e500-876f-11e9-8598-4b34743a546d.png)).

- The receiver receives an email with the subject `New knowledge post with tag [xxx]!`, which on the one hand does not include the title of the post and on the other hand, it only shows a tag of the post (that could have many).

### Proposed solution

- Use a simplified version of the email which shows the TLDR instead of the whole post and include a link to the full article.
- Include the post thumbnail as an attached image so there are no render issues in the different email clients.
- Include more information in the email.

jihonrado:
Fixes #516

### Description of the changeset

We have modified the plain and rich mails to include more information. For the rich email, we have included a new template which shows a preview of the post, along with its thumbnail and related data, with a link to check the full article in the Knowledge Repo.

#### Preview of the rich email

Screenshot 2019-06-05 at 15 38 03

### Test Plan:

Current tests have been modified.

#### Test results

Screenshot 2019-06-05 at 08 37 23

### DIFF
--- a/knowledge_repo/app/utils/emails.py
+++ b/knowledge_repo/app/utils/emails.py
@@ -1,16 +1,16 @@
 """
 This file deals with all of the email functions.
 """
-
+import base64
 import logging
+import re
 
 from flask import render_template, current_app, url_for
 from flask_mail import Message
 
 from ..proxies import db_session
 from ..proxies import current_repo
-from ..models import Email, Subscription, User, Post
-from ..utils.render import render_post
+from ..models import Email, Subscription, User
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


knowledge_repo/app/utils/emails.py
```diff
@@ -88,21 +88,42 @@ def send_subscription_email(post, tag):
...
-    post_text = render_post(post)
-
+    subject = u"New KR post: {}".format(post.title)
+    post_authors = [p.format_name for p in post.authors]
+    post_tags = [t.name for t in post.tags]
+    msg = Message(subject=subject, recipients=default_recipients, bcc=recipients_bcc)
+
+    # Extract bytes from post thumbnail or fallback to the default one
+    if post.thumbnail:
+        reg_search = re.search('data:.*;base64,(.*)', post.thumbnail)
```
serenajiang, line 98: I think we're missing the case where the thumbnail is an image url

jihonrado:
Hi @serenajiang! Sorry for the late response here.

Thanks for taking a look and introducing improvements in #540. I am closing this one to continue the discussion in the new PR.

PR #525, opened 2019-11-16
Add pageview indexes

etr2460:
Description of changeset: 
Addes indexes to speed up queries that take >90% of the load time for the homepage
image

Test Plan: 
CI and tested in Airbnb's environment

@john-bodley @serenajiang @danfrankj

### DIFF
--- a/knowledge_repo/app/migrations/versions/eb39ac93fc39_add_pageview_indexes.py
+++ b/knowledge_repo/app/migrations/versions/eb39ac93fc39_add_pageview_indexes.py
@@ -1,0 +1,26 @@
+"""add_pageview_indexes
+
+Revision ID: eb39ac93fc39
+Revises: d15f6cac07e1
+Create Date: 2019-11-15 15:30:55.010491
+
+"""
+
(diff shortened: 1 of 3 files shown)


knowledge_repo/app/migrations/versions/eb39ac93fc39_add_pageview_indexes.py
```diff
@@ -0,0 +1,26 @@
...
+# revision identifiers, used by Alembic.
+revision = 'eb39ac93fc39'
+down_revision = 'd15f6cac07e1'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.create_index(
```
serenajiang: Since it passes CI, it's probably fine, but in superset we always use `batch_alter_table`/`batch_op` since plain `op` doesn't work with some dialects.

https://alembic.sqlalchemy.org/en/latest/batch.html <- more info

serenajiang: nvm

Issue #531, opened 2020-02-10
ImportError: cannot import name url_encode


kelvins:
I'm facing the following error when trying to run `knowledge-repo==0.8.8`.

`knowledge_repo --repo . runserver --config config.py --port 7000`:

```bash
Traceback (most recent call last):
  File "/home/kelvins/luizalabs/heisenberg/venv/bin/knowledge_repo", line 300, in <module>
    from knowledge_repo.app.deploy import KnowledgeDeployer, get_app_builder
  File "/home/kelvins/luizalabs/heisenberg/venv/lib/python3.6/site-packages/knowledge_repo/app/__init__.py", line 4, in <module>
    from .app import KnowledgeFlask
  File "/home/kelvins/luizalabs/heisenberg/venv/lib/python3.6/site-packages/knowledge_repo/app/app.py", line 20, in <module>
    from werkzeug import url_encode
ImportError: cannot import name 'url_encode'
```

It seems `knowledge-repo` [does not pin the Werkzeug version](https://github.com/airbnb/knowledge-repo/blob/master/knowledge_repo/_version.py#L39) and Werkzeug has released a new version 1.0.0 that breaks the retro compatibility (issue: https://github.com/pallets/werkzeug/issues/1714).

```bash
$ pip show werkzeug

Name: Werkzeug
Version: 1.0.0
Summary: The comprehensive WSGI web application library.
Home-page: https://palletsprojects.com/p/werkzeug/
Author: Armin Ronacher
Author-email: armin.ronacher@active-4.com
License: BSD-3-Clause
Location: /home/kelvins/luizalabs/heisenberg/venv/lib/python3.6/site-packages
Requires: 
Required-by: knowledge-repo, Flask
```

bartaelterman:
Maybe you can suggest a PR? It seems to be here: https://github.com/airbnb/knowledge-repo/blob/master/knowledge_repo/_version.py#L39

wrgoldstein:
👍  I'm having this issue too

gmwill92:
Is there a fix for this?

bartaelterman:
As long as there is no maintainer responding I think your best option is to clone this repository locally, make the modifications from @kelvins [pull request](https://github.com/airbnb/knowledge-repo/pull/532) and install locally.

kelvins:
@gmwill92 another workaround would be to pin the Werkzeug version in your requirements/Pipfile:

```
knowledge-repo[all]
werkzeug==0.16.1
```

[...]

PR #532, opened 2020-02-14
Fix deprecated werkzeug imports removed in version 1.0.0

kelvins:
Fix deprecated Werkzeug imports removed in version 1.0.0:
- `werkzeug.url_encode` -> `werkzeug.urls.url_encode`
- `werkzeug.secure_filename` -> `werkzeug.utils.secure_filename`

This PR solves issue https://github.com/airbnb/knowledge-repo/issues/531.

### DIFF
--- a/knowledge_repo/app/app.py
+++ b/knowledge_repo/app/app.py
@@ -17,7 +17,10 @@ from flask_principal import Principal, identity_loaded, Identity, AnonymousIdent
 from alembic import command
 from alembic.migration import MigrationContext
 from datetime import datetime
-from werkzeug.urls import url_encode
+try:
+    from werkzeug import url_encode
+except ImportError:
+    from werkzeug.urls import url_encode
 
 import knowledge_repo
 from . import routes
--- a/knowledge_repo/app/app.py
+++ b/knowledge_repo/app/app.py
@@ -17,7 +17,7 @@ from flask_principal import Principal, identity_loaded, Identity, AnonymousIdent
 from alembic import command
 from alembic.migration import MigrationContext
 from datetime import datetime
-from werkzeug import url_encode
+from werkzeug.urls import url_encode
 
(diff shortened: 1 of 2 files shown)


bartaelterman:
Hi @kelvins 

First, full disclosure: I am not a maintainer or this project. Just a fan hoping we can push things further because it seems things have stalled here. Hence I will not be able to accept your MR, we'll need input from a maintainer for that.

The fix you proposed is not backwards compatible and hence it will break for people working with older versions of werkzeug. I guess that is also why the Travis build fails (although the output is not entirely clear to me)
I suggest you either wrap the import in a `try/except` block (try the old import first, catch the `ImportError` and then try the new import) or pin the werkzeug version in the _version.py file like:
`werkzeug<1.0` (because jumping to a higher major version could possibly cause more issues in the code.)

kelvins:
Hello @bartaelterman, thanks for your CR and sorry for the delay in answering you.

I didn't understand why the build is failing on Travis, the message is not clear to me.

I have tested the import with older Werkzeug versions and it worked (at least it didn't raise ImportError, but I'm not sure the code behavior is the same). Even so, I think your idea of wrapping the import with `try/except` is great and I have implemented it.

My first thought was to pin the Werkzeug version but I believe we would lose new features/bug fixes from the new Werkzeug versions, so I think wrapping the import with `try/except` would be more appropriate.

I have noticed that the build on Travis is still failing but I'm not sure it is related to the changes in this PR. The PR https://github.com/airbnb/knowledge-repo/pull/533 is also failing even though it has changed only a markdown file.

john-bodley:
@kelvins I've addressed this issue in https://github.com/airbnb/knowledge-repo/pull/536. It's probably preferred to add a constraint on the `werkzeug` package. I verified that version 1.0.0 support all the versions of Python which the Knowledge Repo supports.

PR #540, opened 2020-05-06
Improve subscription emails (revised)

serenajiang:
## Changes
Change subscription emails to contain a summary of the post instead of the entire post. This should fix some recurring issues for reindexing.

Mostly using @jihonrado's changes from #517 + fixing up some issues with the thumbnail image.

## Test Plan
Tested on staging env. Received this email.

image

## Reviewers
@etr2460 @john-bodley

### DIFF
--- a/knowledge_repo/app/templates/email_templates/subscription_email.html
+++ b/knowledge_repo/app/templates/email_templates/subscription_email.html
@@ -1,0 +1,19 @@
+<table border="0" cellpadding="0" cellspacing="0" style="font-size: 13px">
+    <tbody>
+        <tr>
+            <td style="vertical-align: center; text-align: center; width: 150px">
+                <a href="{{ url_for('posts.render', path=page_id, _external=True) }}">
+                    <img style="border: 0; max-width: 140" src="cid:Thumb" />
+                </a>
+            </td>
+            <td style="padding-left: 10px">
+                <p style="font-size: 15px; font-weight: bold">{{ post_title }}</p>
+                <p style="margin: 5px 0">
+                    Author(s): <span style="font-style: italic; font-weight: bold">{{ post_authors|join(', ') }}</span>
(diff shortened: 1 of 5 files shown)


knowledge_repo/app/templates/email_templates/subscription_email.html
```diff
@@ -0,0 +1,19 @@
+<table border="0" cellpadding="0" cellspacing="0" style="font-size: 13px">
```
etr2460, line 1: can we wrap this in <html> and <body> tags as well to follow html conventions?

knowledge_repo/app/utils/emails.py
```diff
@@ -88,16 +89,27 @@ def send_subscription_email(post, tag):
         return
 
     default_recipients = ['knowledge_consumer@notreal.com']
-    subject = u"New KR post: {}".format(post.title)
+    subject = u"New knowledge post with tag [{}]!".format(tag.name)
```
jihonrado, line 92: To clarify why I removed the tag from the subject, I will describe a use case (please remember that a post can have more than one tag assigned):

1. A user is subscribed to tags `tag1` and `tag2`.
1. A post with the tags `tag1` and `tag2` is published.
1. The user described in point `1` receives and email with the subject `New knowledge post with tag [tag1]!`.

I found the following weird things about it:

1. Only one of the two tags was included in the subject and the full list of tags was not included in the email either. 
1. The subject of the email is not relevant at all.

That's why I...:

1. Included the full list of the tags in the new HTML body .
1. Used the post title as email subject for  better readability and identification purposes.
1. Included a prefix in the subject so the user can easily recon that the email was received from the Knowledge Repo. I used `New KR post:` to achieve that but I agree with @serenajiang that acronym could be ambiguous, so I think it is better to use `Knowledge post:`, `New knowledge post:` or similar.

serenajiang: Good points. I'll change it to `New knowledge post: ...`. Thanks!

PR #300, opened 2017-07-07
Add PDF export functionality for knowledge posts

matthewwardrop:
This PR addresses #298 by adding support for exporting knowledge posts to PDFs using `weasyprint`, as well as exposing support for this format in the web app user interface. At this point, the code works, but there are no stylesheets are applied, and so it is very much a work in progress.

Auto-reviewers: @NiharikaRay @matthewwardrop @earthmancash @danfrankj

PR #547, opened 2020-05-26
Support Python >= 3.6

naoyak:
### Description of changeset
Proposing dropping support for Python versions < 3.6 to reduce maintenance overhead and unlock some dependency upgrades (https://github.com/airbnb/knowledge-repo/pull/300#commitcomment-39416214).

Test Plan:

### DIFF
--- a/knowledge_repo/app/routes/debug.py
+++ b/knowledge_repo/app/routes/debug.py
@@ -2,6 +2,8 @@ import sys
 import platform
 import types
 import logging
+from urllib.parse import unquote
+
 import tabulate
 import knowledge_repo
 
@@ -68,7 +70,6 @@ def force_reindex():
 
 @blueprint.route('/debug/views')
 def show_views():
-    import urllib
     output = []
     for rule in current_app.url_map.iter_rules():
         options = {}
@@ -77,7 +78,7 @@ def show_views():
 
         methods = ','.join(rule.methods)
         url = url_for(rule.endpoint, **options)
-        line = urllib.unquote("{:50s} {:20s} {}".format(rule.endpoint, methods, url))
+        line = unquote("{:50s} {:20s} {}".format(rule.endpoint, methods, url))
         output.append(line)
 
     return "<br />".join(sorted(output))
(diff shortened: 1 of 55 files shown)


bulam:
@naoyak Thanks for putting it in! I think this LGTM. @etr2460 Any thoughts?

etr2460:
@john-bodley or @serenajiang, would you mind taking a look at this?

bulam:
@naoyak I think there might be a lot of users using Python 2.7 on their own computers. If they try to add posts to the repo using the older version of the knowledge repo compatible with Python 2.7 do you think there could be compatability issues?

naoyak:
@bulam personally I think it's reasonable to require Python 3 for those who want to use the latest version of KR, given that Python 2 has been EOL'd and most major libraries have dropped support. Perhaps it makes sense to ship this in a new major/minor release along with your styling revamp in #548?

knowledge_repo/app/routes/debug.py
```diff
@@ -77,7 +77,7 @@ def show_views():
 
         methods = ','.join(rule.methods)
         url = url_for(rule.endpoint, **options)
-        line = urllib.unquote("{:50s} {:20s} {}".format(rule.endpoint, methods, url))
+        line = urllib.parse.unquote("{:50s} {:20s} {}".format(rule.endpoint, methods, url))
```
john-bodley, line 80: Maybe use the logic below in terms of importing this.

john-bodley:
@bulam I agree with @naoyak here. It's pretty easy for users to support multiple versions of Python locally via [`pyenv`](https://github.com/pyenv/pyenv).

knowledge_repo/app/routes/debug.py
naoyak: @john-bodley sorry can you clarify? Would it be better to `from urllib.parse import unquote` at the top of the file?

naoyak: Ah I see what you were referring to. Best to keep consistent 👍

PR #549, opened 2020-05-27
Add Mac to Travis environment matrix

naoyak:
## Description of changeset: 
Users running the KR workflow often run it in OSX, so it seems worthwhile to add it to Travis.

cc @bulam @etr2460

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,13 +11,14 @@ jobs:
     - os: linux
       python: 3.8
     - os: osx
+      language: generic
       python: 3.8
       osx_image: xcode11.4
 
 before_install:
   # We do this conditionally because it saves us some downloading if the
   # version is the same.
-  - if [ $TRAVIS_OS_NAME == 'osx' ]; then
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
       wget https://repo.continuum.io/miniconda/Miniconda3-latest-MacOSX-x86_64.sh -O miniconda.sh;
     else if [[ "$TRAVIS_PYTHON_VERSION" == "2.7" ]]; then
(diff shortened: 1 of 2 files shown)


naoyak:
@bulam I've gotten the Mac build working on Travis. PTAL!

.travis.yml
```diff
@@ -1,17 +1,27 @@
...
+      python: 2.7
+    - os: linux
+      python: 3.6
+    - os: linux
+      python: 3.7
+    - os: linux
+      python: 3.8
+    - os: osx
+      language: generic
+      python: 3.6
```
bulam, line 15: Any reason we're doing only python 3.6 on OSX?

naoyak: The Mac tests kind of take a long time and I think it's ok for now to set this up for 3.6. Usually it should work on 3.7 and 3.8 as well as long as there aren't any missing pip wheel dependencies

bulam: 👍

etr2460:
Sorry, why did we need to add Mac to the Travis build matrix? I don't think anyone is hosting the knowledge repo server on a Mac, so it seems like this shouldn't be necessary

naoyak:
@etr2460 you're right that there probably isn't anyone hosting the actual server on Mac. I was thinking more about the user workflow - after all the CI setup already includes Appveyor for Windows, and some users have reported issues running on Windows.

Maybe it makes the most sense to have certain parts of the codebase covered by tests running on Linux only, while the CLI / user-facing components have coverage on Linux/Mac/Windows. I imagine this could be done pretty easily, but the entire test suite is pretty small for now so it wouldn't have a big impact either.

PR #552, opened 2020-05-27
Header parsing

### LINKED ISSUE #483: using emojis in markdown cells raises errors about headers

When there is an emoji in a markdown cell in my document, I get the following error:
```
RuntimeError: Knowledge post is missing required headers {'title', 'tldr', 'created_at', 'authors'}. Please rerun this operation in interactive mode, or add headers manually to the post source file.
```

This error doesn't make sense as the headers are not missing. 

simple example created with `knowledge_repo create ipynb test.ipynb` then pasting an emoji in

image

The first markdown block is changed to this:
```
*NOTE: In the TL,DR, optimize for **clarity** and **comprehensiveness**. The goal is to convey the post with the least amount of friction, especially since ipython/beakers require much more scrolling than blog posts. Make the reader get a correct understanding of the post's takeaway, and the points supporting that takeaway without having to strain through paragraphs and tons of prose. Bullet points are great here, but are up to you. Try to avoid academic paper style abstracts.*

 - Having a specific title will help avoid having someone browse posts and only finding vague, similar sounding titles
 - Having an itemized, short, and clear tl,dr will help readers understand your content
 - Setting the reader's context with a motivation section makes someone understand how to judge your choices
 - Visualizations that can stand alone, via legends, labels, and captions are more understandable and powerful
 - 😊
```

### LINKED ISSUE #489: Post header parsing fails with windows-style (CRLF) line endings

Hi and thank you very much for this amazing project. I tried it out today ran into errors. After several attempts to resolve them (and starting to write issues), it boiled down to this:

### Problem

When files are comitted with windows-style (CRLF) endings, the web app is unable to render (or even find) the post. The previewer (`knowledge_repo --repo <repo-path> preview <post-path>`) fails when looking for the YAML header with the following error:

```
ValueError: YAML header is missing. Please ensure that the top of your post has a header of the following form:

---
title: "This is a Knowledge Post title, quoted so we can use special characters like ':'"
authors:
- sally_smarts
- wesly_wisdom
tags:
- knowledge
- example
created_at: 2016-06-29
updated_at: 2016-06-30
thumbnail: 2
tldr: |
    You can write any markdown you want here (the '|' character makes this an escaped section)

    * bullet
    * bullet

    You can even write arbitrary html (html is valid markdown)
    <table>
    <tr>
    <th>hi</th>
    <th>I'm </th>
    </tr>
    <tr>
    <td>a</td>
    <td>table</td>
    </tr>
    </table>
---
```

which misleadingly lead me to try and modify the header. Interestingly enough, this happened only with a defined remote for the git repo. When I added a post with no remote defined for the repo, everything worked as expected. 

### Solution

As it turned out, the parser expects unix-style (LF) line endings and breaks on windows-style (CRLF) line endings. The recommended way to use git on windows is "checkout as windows-style, commit unix-style" which corresponds to `git config core.autocrlf true`. Manually changing the line endings (using PyCharm) and enforcing unix-style line endings everywhere (`git config core.autocrlf input` and adding this [.gitattributes](https://github.com/airbnb/knowledge-repo/files/2728611/gitattributes.txt) to the repo solved the issue. (I did not retry both ways independently.)

For some reason this was not necessary with a local-only repo. This would make sense, if files are committed LF locally by default, but checked out as CRLF when pulling from the remote repo. Does this happen? 

Either way, as it seems, CRLF line endings pose a problem and I wanted to tell you about it ;-)

naoyak:
## Description of changeset: 
Closes #503, closes #490
Fixes #489, fixes #483 

Combining the YAML header detection issues addressed by #489 and #483.

Tested by processing a known good Jupyter notebook locally through `knowledge_repo add`.

### DIFF
--- a/knowledge_repo/app/templates/markdown-presentation.html
+++ b/knowledge_repo/app/templates/markdown-presentation.html
@@ -35,7 +35,7 @@
   });
 </script>
 <script type="text/javascript"
-  src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.7/MathJax.js">
+  src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.8/MathJax.js">
 </script>
 <script>
   $(".btn-presentation").addClass("btn-active");
--- a/knowledge_repo/app/templates/markdown-rendered.html
+++ b/knowledge_repo/app/templates/markdown-rendered.html
@@ -108,7 +108,7 @@
   });
 </script>
 <script type="text/javascript"
-  src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.7/MathJax.js">
+  src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.8/MathJax.js">
 </script>
 
 <script>
--- a/knowledge_repo/app/templates/post_editor_base.html
+++ b/knowledge_repo/app/templates/post_editor_base.html
@@ -560,7 +560,7 @@ function parseDate(ds, error_name){
   });
 </script>
 <script type="text/javascript"
-  src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.7/MathJax.js">
+  src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.8/MathJax.js">
 </script>
 
 {% endblock %}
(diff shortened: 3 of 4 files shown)


bulam:
@naoyak Did you also test on a knowledge post containing an emoji just to make sure it's working?

naoyak:
> @naoyak Did you also test on a knowledge post containing an emoji just to make sure it's working?

@bulam just tested. seems to work!
Screen Shot 2020-05-28 at 7 25 26 PM

bulam:
Nice! Allows emojis in headers as well, not just in the body!

PR #557, opened 2020-05-30
Handle nested and list identifier keys for OAuth response

### LINKED ISSUE #468: Bug in Auth_providers that won't let you sign in the Knowledge Repo App

I would like to bring your attention to an issue I figured out with respect to the knowledge repo app:

I used Github OAuth authentication to allow different users sign in the app by redirecting them to their Github account. When I pressed the sign in button, although I would get redirected to my Github account, I didn't get redirected and signed in the app. After some research, I managed to solve the problem by making a small change in "..../knowledge_repo/app/auth_providers/oauth2.py", in line 32, where I switched 
'identifier': ['email', 'login'] 
to 
'identifier': ['login'].

The problem seemed to be caused by the fact that function "extract_from_dict" expected to find email and login in a nested form, while that's not the case for the Github OAuth authentication as can be seen here (https://developer.github.com/v3/users/#get-the-authenticated-user):
image

Changing line 32 bypasses this problem by only looking for the one of the two fields, but a more robust solution would be to update the extract_from_dict function in a way that it will expect the requested fields in both nested and not nested forms, so that all kinds of authenticators can be utilized, independently of the exact way with which they store users information.

### LINKED ISSUE #510: Auth Provider with GitHub
I was try to GitHub as auth provider, when updated auth provider in config_defaults.py, changes not apply when I start server, can you please tell me what I am missing.

naoyak:
## Description of changeset: 
Fix #468, fix #510.

## Test Plan: 
- [x] Tested in development with GitHub OAuth

### DIFF
--- a/knowledge_repo/app/auth_providers/oauth2.py
+++ b/knowledge_repo/app/auth_providers/oauth2.py
@@ -152,11 +152,17 @@ class OAuth2Provider(KnowledgeAuthProvider):
     def extract_user_from_api(self):
 
         def extract_from_dict(d, key):
-            if isinstance(key, (list, tuple)):
+            if isinstance(key, tuple):
                 if len(key) == 1:
                     key = key[0]
                 else:
                     return extract_from_dict(d[key[0]], key[1:])
+            if isinstance(key, list):
+                for _, k in enumerate(key):
+                    try:
+                        return extract_from_dict(d, k)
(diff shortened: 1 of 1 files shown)


naoyak:
I've gotten the OAuth login part working again at least with GitHub. @bulam

PR #571, opened 2020-07-01
Removing read of knowledge repo config file off default branch of master

### LINKED ISSUE #569: Bug in support for non-master default branches for repos
Hello! My team has made the decision to move our tools away from using `master`, instead opting for `main` as the new default branch pattern. 

As I understand the error message I'm getting using the `knowledge_repo` tool with the `main` branch for the git repository knowledge repo, the code flow looks something like this when running `knowledge_repo add ...`

1. [Set a default published branch to "master"](https://github.com/airbnb/knowledge-repo/blob/f026ad2afea14e0ffc91f1aa0eaedcdc72c63167/knowledge_repo/repositories/gitrepository.py#L58)
1. Read the config of the repository for updated values [using the git_read method](https://github.com/airbnb/knowledge-repo/blob/f026ad2afea14e0ffc91f1aa0eaedcdc72c63167/knowledge_repo/repositories/gitrepository.py#L79) 
1. Break, as the git reader requires a valid `published_branch` value before it can set it to the repo's config override here [in the reader](https://github.com/airbnb/knowledge-repo/blob/f026ad2afea14e0ffc91f1aa0eaedcdc72c63167/knowledge_repo/repositories/gitrepository.py#L182)
1. The issue is raised [as a git.GitError](https://github.com/airbnb/knowledge-repo/blob/f026ad2afea14e0ffc91f1aa0eaedcdc72c63167/scripts/knowledge_repo#L95) and suppressed, where I'm seeing a different issue later as the `repo` object is `None`

Thanks for any support here!

I'll try my hand at a pull request as I can

e-m-albright:
Fixes #569 

Description of changeset: 

It looks like you're reading the state of the config file of the knowledge repo off of the target branch which is hardcoded to default as master before there's any room to update the preference of what branch to use.

As a simple fix perhaps it's best to read the state of the config for the knowledge repo of whatever the active branch is instead? This alleviates the breakage when there is no master branch to check the file out from in the git of the knowledge repo.

Test Plan: 

Tried pytest, tried `bash run_tests.sh`, neither worked. I'm not sure if the issues are environmental and I should be using another tool to aid with them.

Would love to know how to help further! Figure this is at least a good starting point to help me learn more about your systems.

### DIFF
--- a/knowledge_repo/repositories/gitrepository.py
+++ b/knowledge_repo/repositories/gitrepository.py
@@ -76,7 +76,7 @@ class GitKnowledgeRepository(KnowledgeRepository):
         if config.startswith('git:///'):
             assert config.endswith('.yml'), "In-repository configuration must be a YAML file."
             try:
-                self.config.update(yaml.safe_load(self.git_read(config.replace('git:///', ''))))
+                self.config.update(yaml.safe_load(config.replace('git:///', '')))
             except KeyError:
                 logger.warning("Repository missing configuration file: {}".format(config))
         else:


bulam:
Thanks for the contribution! Will take a look when I get a chance :)

PR #583, opened 2020-10-20
Allow users to load their own Gunicorn config

dmaljovec:
What?
=====

* Port relevant sections of code from `gunicorn.app.base.Application` into `knowledge-repo.app.deploy.gunicorn.GunicornDeployer` to use the `--config` supplied by the user.

Why?
=====

* We'd like to instrument our deployment of Knowledge-Repo with some observability metrics and being able to customize our gunicorn deployment is the way we do that with other apps right now.

This builds off of https://github.com/airbnb/knowledge-repo/pull/437 where a user can supply their own arguments. Unfortunately, while I can see that it does indeed bring in my `GUNICORN_CMD_ARGS='--config gunicorn.conf.py` it does not actually load it up. I stole some code from [gunicorn's source](https://github.com/benoitc/gunicorn/blob/ee685e197b3f7cf899dc7d6e0688ff169e9d10df/gunicorn/app/base.py#L173-L180) and adapted it gently to work in the knowledge-repo use case. Please let me know if this is something you think is worth supporting and if you need anything additional to make this go (tests, more explanation, a usage example, etc.).

### DIFF
--- a/knowledge_repo/app/deploy/gunicorn.py
+++ b/knowledge_repo/app/deploy/gunicorn.py
@@ -9,6 +9,13 @@ from gunicorn.app.base import BaseApplication
 
 from .common import KnowledgeDeployer
 
+import importlib.machinery
+import importlib.util
+import os
+import sys
+import traceback
+import warnings
+
 
 class GunicornDeployer(BaseApplication, KnowledgeDeployer):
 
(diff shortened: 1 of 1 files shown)


dmaljovec:
I tested this in a test deployment and it appears to do what I want, FWIW.

naoyak (approved):
Thanks!

naoyak:
@dmaljovec Sorry about the long delay. Is there a way you could add a test?

dmaljovec:
> @dmaljovec Sorry about the long delay. Is there a way you could add a test?

Yeah, sure. I'll try to have a look before the end of the week.

dmaljovec:
Sorry, I took a brief look and it looks like the `GunicornDeployer` is not under test yet (actually, it looks like none of the `KnowledgeDeployer` subclasses are), so it might take more effort than I first thought to write some unit tests around this. I haven't thought much about it and maybe it is easier than I am making it out to be, but it might be some time before I get a chance to think about testing this component in isolation. I was hoping to just plug in some minor modifications of existing tests.

PR #593, opened 2021-05-04
Improve connectivity check for git repositories

gthomas-slack:
The `GitKnowledgeRepository` class has an internal function `__remote_available` which is used to check if there is network connectivity to the remote git repo before attempting to execute a `fetch` or `push` to the repo.

This works for most `http` and `ssh` git repos , however it does not work in the case of an `ssh` repo where there is no direct connectivity to the ssh server, but rather there is an ssh configuration which proxies the connection through an intermediary server.

Currently with this kind of repository setup Knowledge Repo will abort fetches, and raise an exception for pushes even though there is actually no connectivity issue.

This PR removes the `__remote_available` function, and replaces it with some try-except blocks which should achieve the same intention, but also work for these types of ssh repos.

In addition the try-except is more helpful to the user, as it will notify them not only of connectivity issues, but also other issues such as permissions issues, for example:
```bash
RuntimeError: Failed to push to remote repository origin (git@github.com:some-repo/we-dont-have-access-to.git). Please check the following error, and then try again:

stderr: 'fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.'
```

### DIFF
--- a/knowledge_repo/repositories/gitrepository.py
+++ b/knowledge_repo/repositories/gitrepository.py
@@ -134,12 +134,15 @@ class GitKnowledgeRepository(KnowledgeRepository):
         branch = branch or self.config.published_branch
         if not self.git_has_remote:
             return
-        if not self.__remote_available:
-            logger.warning("Cannot connect to remote repository hosted on {}. Continuing locally with potentially outdated code.".format(
+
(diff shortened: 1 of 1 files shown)


naoyak (approved):
Thanks!

PR #647, opened 2022-04-01
remove auto format from vscode setting

JJJ000:
Description of changeset: 
remove auto format from vscode setting

Test Plan: 
[x]CI

@csharplus


csharplus (approved):
LGTM.

PR #662, opened 2022-05-01
add doc workflow

JJJ000:
Description of changeset: 

add doc workflow in github action

Test Plan: 
[x]CI

@csharplus @bulam

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -16,7 +16,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install -U pip
-          python -m pip install -r .github/workflows/requirements.txt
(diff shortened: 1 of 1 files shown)


csharplus (approved):
LGTM.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,35 @@
+name: Docs
+on: [push, release]
+
+jobs:
+  notebooks:
+    name: "Build the notebooks for the docs"
+    runs-on: ubuntu-latest
```
csharplus, line 7: You may want to pick a fixed `ubuntu` version to make build more stable, as new versions may easily break the build.

csharplus:
@JJJ000 is this PR still needed, or it could be closed?

PR #668, opened 2022-05-10
remove unintentional wildcard from URL checks

adhintz:
the previous code would unintentionally match https://docsXgoogle.com/...


JJJ000 (approved):

adhintz:
Thanks for the approval! It looks like only those with write access to this repository can merge pull requests.

PR #669, opened 2022-05-23
fix package dependency to support local server

JJJ000:
Description of changeset: 
fix package dependency to support local server

Test Plan: 
[x]ci

@csharplus

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -4,7 +4,7 @@ LABEL description="knowledge-repo service"
 
 # Define some build time variables
 ARG PIP=pip3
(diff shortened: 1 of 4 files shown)


docker/Dockerfile
```diff
@@ -4,7 +4,7 @@ LABEL description="knowledge-repo service"
 
 # Define some build time variables
 ARG PIP=pip3
-ARG VERSION=0.9.1
+ARG VERSION=0.9.0
```
csharplus, line 7: Avoid reducing the version number?

csharplus (approved):
LGTM

PR #673, opened 2022-06-12
fix index and db bugs

JJJ000:
Description of changeset: 
1.  index_failure when service start
2. PostgreSQL close unexpectedly the connection issue

Test Plan: 
local build

@csharplus

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -22,4 +22,4 @@ requests
 weasyprint
 jinja2>=2.7,<=3.0.3
 werkzeug>=1.0,<=2.0.3
-multiprepo
+multiprocess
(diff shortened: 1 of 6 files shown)


requirements.txt
```diff
@@ -22,3 +22,4 @@ requests
 weasyprint
 jinja2>=2.7,<=3.0.3
 werkzeug>=1.0,<=2.0.3
+multiprepo
```
csharplus, line 25: Is this the correct library name?  I searched for it and could not find it.  I found the following instead: https://pypi.org/project/multiprocess/

JJJ000: @csharplus yeah, you are right, I past the wrong package name

csharplus (approved):
LGTM

PR #678, opened 2022-10-26
knowledge v2

JJJ000:
Description of changeset: 

Checkin in This script is intended to created a seperate v2 process wroking in parallel with existing knowledge_repo
At this stage, v2 behavior will 100% match with exsiting worflow

Test Plan: 
N/A

Reviewers:
@csharplus


csharplus (approved):
LGTM.

PR #684, opened 2022-11-22
[kp2]add volumes  and auto sync in docker compose

JJJ000:
Description of changeset: 
add volumes  and auto sync in docker compose

Test Plan: 

docker-compose up -d
docker exec -it [container_name]

Reviewers:
@csharplus


csharplus (approved):
LGTM.

PR #703, opened 2022-12-31
Avoid printing-out exception errors when REVISION or UUID files are not found

csharplus:
Description of changeset:

When creating a new post with the `knowledge_repo` tool, missing REVISION or UUID files are expected, and should not lead to exception messages.

Test Plan:
CI

Reviewers:
@JJJ000 @mengting1010


JJJ000 (approved):
Just curious, we are setting up git repo instead of folder repo. Error message might generate due to incorrect repo setting.

PR #710, opened 2023-01-04
update views and metadata

mengting1010:
Description of changeset:
- change from distinct views to all views
- hide created_at and updated_at fields for ipynb post editor

Test Plan:
local dev

Reviewers:
@csharplus @JJJ000


csharplus (approved):
LGTM.

PR #725, opened 2023-01-20
update notion db id

mengting1010:
Description of changeset:
as title

Test Plan:
local dev

Reviewers:
@csharplus @JJJ000


knowledge_repo/utils/notion.py
```diff
@@ -65,7 +65,7 @@ def create_page(notion_client, params):
         notion_client.pages.create(
             parent={
                 "type": "database_id",
-                "database_id": "85db4c309b6d4d00a6a1df19d56496b6",
```
JJJ000, line 68: we probably need to add it to config

mengting1010: Updated, PTAL again. Thanks!

JJJ000 (approved):
nice nice nice

PR #726, opened 2023-01-20
Remove the # symbol before tags to avoid unnecessary confusion

csharplus:
Description of changeset:
Currently all tags are displayed with "#" prefix, which is unnecessary, and also could be confused with some real tag symbols, e.g., "C#".

Test Plan:
CI

Reviewers:
@JJJ000 @mengting1010


csharplus:
@JJJ000 @mengting1010 PTAL

JJJ000 (approved):

PR #744, opened 2023-02-08
[kp]fix install

JJJ000:
Description of changeset:
fix install

Test Plan:
local

Reviewers:
@csharplus @mengting1010


csharplus (approved):
LGTM.

PR #760, opened 2023-02-28
Fix Plotly in Rmarkdown

calabozo:
Description of changeset:
When generating reports from Rmarkdown files, the Rscript command will include the javascript code generated by Plotly inside a tags called {=html}. This is further ignored by knowledge_repo which includes that text directly in the final repo. Wiith this patch we remove the {=html} and the plotly graphs are properly displayed in the final generated HTML.

Test Plan:

Reviewers:


csharplus (approved):
LGTM.

PR #781, opened 2025-07-16
Remove s3path unused dependency

karyotakisg:
As part of our ongoing research on Python dependency management we noticed a potential improvement in your project’s dependency management.

Specifically, the direct dependency `s3path` is specified as a requirement in the `requirements.txt` and `_version.py` file, when in reality it is not needed.

`s3path` is unused since 04c1626, where its import was removed.

Hope this is helpful!

Best regards


strangeleo02 (approved):
LGTM. Verified s3path isn't used anywhere else in the codebase, so it's safe to drop from requirements.txt and -version.py. Good cleanup.

